PR #9, opened 2020-08-07
Add Kustomization patch support to helm3 post render hook, resolves #1

### LINKED ISSUE #1: Post Rendering Patches
It should be possible to provide post rendering patches which will make last mile configurations using
[`post rendering`](https://helm.sh/docs/topics/advanced/#post-rendering) option of Helm. `spec.forProvider.patchesFrom`
array will be used to specify patch definitions satisfying [kustomizes patchTransformer interface](https://kubernetes-sigs.github.io/kustomize/api-reference/kustomization/patches/).

Example:

```
patches:
- patch: |-
    - op: replace
      path: /some/existing/path
      value: new value
  target:
    kind: MyKind
    labelSelector: "env=dev"
- patch: |-
    - op: add
      path: /spec/template/spec/nodeSelector
      value:
        node.size: really-big
        aws.az: us-west-2a
  target:
    kind: Deployment
    labelSelector: "env=dev"
```

More context: https://github.com/crossplane/crossplane/pull/1642#issuecomment-659540625
Design doc: https://github.com/crossplane/crossplane/blob/master/design/one-pager-helm-provider.md#crossplane-helm-provider

lukeweber:


pkg/clients/helm/kustomize.go
```diff
@@ -0,0 +1,80 @@
...
+
+// KustomizationRender Implements helm PostRenderer interface
+type KustomizationRender struct {
+	patches []types.Patch
+	logger  logging.Logger
+}
+
+// Run runs a set of Kustomize patches against yaml input and returns the patched content.
+func (kr KustomizationRender) Run(renderedManifests *bytes.Buffer) (modifiedManifests *bytes.Buffer, err error) {
+	d, err := ioutil.TempDir("", "helm-post-render")
```
turkenh, line 29: nit: "helm-post-render" => make a const

go.mod
```diff
@@ -8,8 +8,8 @@ require (
 	github.com/crossplane/crossplane-tools v0.0.0-20200412230150-efd0edd4565b
 	github.com/google/go-cmp v0.4.0
 	github.com/pkg/errors v0.9.1
-	golang.org/x/tools v0.0.0-20191018212557-ed542cd5b28a
-	golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543
+	golang.org/x/tools v0.0.0-20191018212557-ed542cd5b28a // indirect
+	golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 // indirect
```
turkenh, line 12: `make build` deletes these two lines on my side.
Could you check and update go.mod if it is same on your side.

lukeweber:
Follow up ticket to upgrade on patch updates - https://github.com/crossplane-contrib/provider-helm/issues/11

PR #12, opened 2020-08-11
Trigger upgrade when kustomize patches are modified

### LINKED ISSUE #11: Trigger helm release upgrade when PatchesFrom are changed
In PatchesFrom we don't detect changes from the related configmaps and secrets.

We need to detect if a change has occurred and trigger a helm upgrade per the design [triggering-an-upgrade](https://github.com/crossplane/crossplane/blob/master/design/one-pager-helm-provider.md#triggering-a-helm-upgrade)

lukeweber:
Add logic to get shaSum of all computed post render hook patches and trigger helm upgrade when there is a change.

Resolves #11

### DIFF
--- a/apis/v1alpha1/types.go
+++ b/apis/v1alpha1/types.go
@@ -88,6 +88,7 @@ type ReleaseSpec struct {
 type ReleaseStatus struct {
 	runtimev1alpha1.ResourceStatus `json:",inline"`
 	AtProvider                     ReleaseObservation `json:"atProvider,omitempty"`
+	PatchesSha                     string             `json:"patchesSha,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/config/crd/helm.crossplane.io_releases.yaml
+++ b/config/crd/helm.crossplane.io_releases.yaml
@@ -342,6 +342,8 @@ spec:
                 - type
                 type: object
               type: array
+            patchesSha:
+              type: string
           type: object
       required:
       - spec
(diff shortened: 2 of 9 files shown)


lukeweber:
WIP - Would like some quick feedback before I add tests.
The design called for adding UID + revision of each patch, but all patches are collected anyways and it seems easier and less error prone to just get a hash of the set of patches.
The remote observed release isn't actually a CRD as I understand it so it seems like patching the local release object with the sha of the local patches after and update/create has happened.

turkenh:
> The design called for adding UID + revision of each patch, but all patches are collected anyways and it seems easier and less error prone to just get a hash of the set of patches.

Yes, this sounds like a better solution.

> The remote observed release isn't actually a CRD as I understand it so it seems like patching the local release object with the sha of the local patches after and update/create has happened.

My only concern would be possible edge cases like, e.g. upgrade succeeded but annotating release object failed. By putting annotation on k8s `Release` object, we are kind of storing actual state on CR. This is why I was trying to store patch information of remote observed helm release and best place I found was [Chart.Metadata.Annotations](https://github.com/helm/helm/blob/3d64c6bb5495d4e4426c27b181300fff45f95ff0/pkg/chart/metadata.go#L58) which is also not great. 
Stating this, I am also fine with storing sha on `Release` CR as well as long as we consider/handle possible edge cases.

lukeweber:
@kferrone - When this PR lands we should have support for basic patching and detection of updates for helm releases.

turkenh (approved):
lgtm!

PR #15, opened 2020-08-22
If version is set to "" - pull latest version from the repo and update the spec.

ktintc:
I kept version as required field in the spec to make sure it's explicitly set to "" if we want to pull the latest chart version (vs accidentally forgot to set the version).


pkg/clients/helm/client.go
```diff
@@ -107,26 +112,100 @@ func NewClient(log logging.Logger, config *rest.Config, namespace string) (Clien
 	}, nil
 }
 
-func (hc *client) pullAndLoadChart(repo, name, version, username, password string) (*chart.Chart, error) {
+func getVersion(chartFileName string) string {
+	versionIndex := strings.LastIndex(chartFileName, "-")
```
turkenh, line 116: nit: would be more explicit/readable with `strings.Split(chartFileName, "-")` followed by a `strings.TrimSuffix(strParts[1], ".tgz")`. 
Otherwise some comments could be helpful.

pkg/clients/helm/client.go
```diff
@@ -107,26 +112,100 @@ func NewClient(log logging.Logger, config *rest.Config, namespace string) (Clien
...
+		for _, f := range files {
+			fileNames = append(fileNames, f.Name())
+		}
+		return "", errors.Errorf(errUnexpectedDirContentTmpl, strings.Join(fileNames, ","))
+	}
+	return files[0].Name(), nil
+}
+
+// Pulls latest chart version. Returns absolute chartFilePath or error.
+// Side effect: sets version of the latest pulled chart to the spec.
```
turkenh, line 136: I would prefer explicit return param instead of implicit `side effect` with reference values.

pkg/controller/release.go
```diff
@@ -237,14 +245,22 @@ func (e *helmExternal) Update(ctx context.Context, mg resource.Managed) (managed
...
 	rel, err := e.helm.Upgrade(meta.GetExternalName(cr), cd, cv, p)
 	if err != nil {
 		return managed.ExternalUpdate{}, errors.Wrap(err, errFailedToUpgrade)
 	}
+
 	if rel == nil {
 		return managed.ExternalUpdate{}, errors.New(errLastReleaseIsNil)
 	}
 
+	if updateToLatest {
```
turkenh, line 258: Instead of passing chart spec as reference and updating version implicitly, or, returning version as yet another parameter explicitly, we could just get the version from release at this point as:

```
cr.Spec.ForProvider.Chart.Version = rel.Chart.Metadata.Version
```  

Then, we even wouldn't need to infer version from filename etc.

One possible problem with this would be only to be able to get/set version after a successful `helm install/upgrade`, but it looks like this is already the case and solution for that seems like to separate `pullAndLoad` as yet another function in helm client interface instead of embedding into `Install` and `Upgrade` functions. Then we could get version from chart metadata instead of release metadata as `chart.Metadata.Version` after calling `PullAndLoad` before Install/Upgrade.

pkg/clients/helm/chartdefinition.go
```diff
@@ -1,10 +1,12 @@
...
-	Repository string
-	Name       string
-	Version    string
-	RepoUser   string
-	RepoPass   string
+	RepoUser string
+	RepoPass string
+
+	// Need to keep reference to be able to update version
+	ChartSpec *v1alpha1.ChartSpec
```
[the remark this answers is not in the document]
ktintc: I agree, let's clean this up.
I don't see the benefit of having Chartdef structure at all: extra fields copying making it harder to follow the logic.

pkg/clients/helm/client.go
ktintc: I agree completely. i just tried to keep the interface untouched initially, gladly updated :-)

pkg/controller/release.go
ktintc: Good point! Updated the interface and use the version from the chart.

pkg/clients/helm/client.go
```diff
@@ -107,26 +115,95 @@ func NewClient(log logging.Logger, config *rest.Config, namespace string) (Clien
...
 }
 
-func (hc *client) pullAndLoadChart(repo, name, version, username, password string) (*chart.Chart, error) {
+/*
+func getVersion(chartFileName string) string {
+	versionIndex := strings.LastIndex(chartFileName, "-")
+	return chartFileName[versionIndex+1 : len(chartFileName)-4]
+}
+
+*/
```
turkenh, line 124: please remove commented code

pkg/clients/helm/userinfo.go
```diff
@@ -0,0 +1,7 @@
+package helm
+
+// UserInfo keeps auth information to access a Helm Chart
+type UserInfo struct {
```
turkenh, line 4: nit: I would suggest naming this struct as `RepoCreds` or `PullCreds` with fields `Username` & `Password`

pkg/controller/chart.go
```diff
@@ -39,32 +39,29 @@ const (
 	errChartPullSecretMissingPassword  = "password missing in chart pull secret"
 )
 
-func chartDefFromSpec(ctx context.Context, kube client.Client, spec v1alpha1.ChartSpec) (helmClient.ChartDefinition, error) {
+func userInfoFromSecret(ctx context.Context, kube client.Client, secretRef runtimev1alpha1.SecretReference) (*helmClient.UserInfo, error) {
```
turkenh, line 42: nit: same as above, `repoCredsFromSecret`

turkenh (approved):
Thank you, looking good.
Just added couple of nits.

pkg/clients/helm/userinfo.go
ktintc: sounds good!

pkg/clients/helm/client.go
ktintc: oops, forgot to push the latest :)

pkg/controller/chart.go
ktintc: done

PR #16, opened 2020-08-26
Implement optional rollback in case of failure

### LINKED ISSUE #3: Optional rollback in case of failure

### LINKED ISSUE #17: ValueFrom field does not meet Kubernetes API conventions
The json tag seems to be forgotten as `ValueFrom` where it should've been `valueFrom`: https://github.com/crossplane-contrib/provider-helm/blob/d313bcf/apis/v1alpha1/types.go#L56

turkenh:
Resolves #3 

Fixes #17


apis/v1alpha1/types.go
```diff
@@ -76,19 +76,23 @@ type ReleaseParameters struct {
...
 	ForProvider                  ReleaseParameters `json:"forProvider"`
+	RollbackLimit                *int32            `json:"rollbackLimit,omitempty"`
 }
 
 // A ReleaseStatus represents the observed state of a Release.
 type ReleaseStatus struct {
 	runtimev1alpha1.ResourceStatus `json:",inline"`
 	AtProvider                     ReleaseObservation `json:"atProvider,omitempty"`
 	PatchesSha                     string             `json:"patchesSha,omitempty"`
+	Failed                         int32              `json:"failed,omitempty"`
```
ktintc, line 94: maybe NumFailed or NumFailures or FailureCounter.
To point out it's purpose is to track number of failures, otherwise looks like a bool to me (especially next to Synced field).

examples/sample/release.yaml
```diff
@@ -3,29 +3,33 @@ kind: Release
...
-          key: values.yaml
-          name: default-vals
-          namespace: wordpress
-          optional: false
-      - secretKeyRef:
-          key: svalues.yaml
-          name: svals
-          namespace: wordpress
-          optional: false
+#    valuesFrom:
```
ktintc, line 22: forgotten to cleanup commented lines?

pkg/controller/observe.go
```diff
@@ -40,12 +40,20 @@ func generateObservation(in *release.Release) v1alpha1.ReleaseObservation {
...
 		o.State = relInfo.Status
 		o.ReleaseDescription = relInfo.Description
+		o.Revision = in.Version
 	}
 	return o
 }
 
 // isUpToDate checks whether desired spec up to date with the observed state for a given release
-func isUpToDate(ctx context.Context, kube client.Client, in *v1alpha1.ReleaseParameters, observed *release.Release, s v1alpha1.ReleaseStatus) (bool, error) {
+func isUpToDate(ctx context.Context, kube client.Client, in *v1alpha1.ReleaseParameters, observed *release.Release, s v1alpha1.ReleaseStatus) (bool, error) { // nolint:gocyclo
```
ktintc, line 49: maybe it's time to increase a limit in linter config since we're adding this after every meaningful function here.

pkg/controller/observe.go
```diff
@@ -40,12 +40,20 @@ func generateObservation(in *release.Release) v1alpha1.ReleaseObservation {
...
+	if observed.Info != nil &&
```
ktintc, line 50: func isPending (?)

pkg/controller/release.go
```diff
@@ -164,20 +164,29 @@ func (e *helmExternal) Observe(ctx context.Context, mg resource.Managed) (manage
...
 	cr.Status.AtProvider = generateObservation(rel)
-	u, err := isUpToDate(ctx, e.localKube, &cr.Spec.ForProvider, rel, cr.Status)
+	s, err := isUpToDate(ctx, e.localKube, &cr.Spec.ForProvider, rel, cr.Status)
 	if err != nil {
 		return managed.ExternalObservation{}, errors.Wrap(err, errFailedToCheckIfUpToDate)
 	}
+	cr.Status.Synced = s
+	shouldCallUpdate := !s
+
+	if cr.Status.AtProvider.State == release.StatusDeployed && s {
```
ktintc, line 174: func rollbackLimitReached or similar would help reader to follow the logic.

pkg/controller/release.go
```diff
@@ -164,20 +164,29 @@ func (e *helmExternal) Observe(ctx context.Context, mg resource.Managed) (manage
...
+		cr.Status.Failed = 0
+	} else if cr.Status.AtProvider.State == release.StatusFailed &&
+		cr.Spec.RollbackLimit != nil && cr.Status.Failed < *cr.Spec.RollbackLimit {
+		shouldCallUpdate = true
+	}
 
 	return managed.ExternalObservation{
 		ResourceExists:   true,
-		ResourceUpToDate: u,
+		ResourceUpToDate: !shouldCallUpdate,
```
ktintc, line 183: !cr.status.synced && !rollbackLimitReached would be easier to understand what condition actually triggers the update.

pkg/controller/release.go
```diff
@@ -164,20 +164,29 @@ func (e *helmExternal) Observe(ctx context.Context, mg resource.Managed) (manage
...
 	}, nil
 }
 
-type chartAction func(release string, chart *chart.Chart, vals map[string]interface{}, patches []ktype.Patch) (*release.Release, error)
+type deployFunc func(release string, chart *chart.Chart, vals map[string]interface{}, patches []ktype.Patch) (*release.Release, error)
```
ktintc, line 187: helm refers to these (install/upgrade etc) as `actions`
https://github.com/helm/helm/tree/d6708667da5c79ab4a093ea7473e96b93be5f1f7/pkg/action

pkg/controller/release.go
```diff
@@ -240,8 +249,26 @@ func (e *helmExternal) Update(ctx context.Context, mg resource.Managed) (managed
 		return managed.ExternalUpdate{}, errors.New(errNotRelease)
 	}
 
+	if cr.Spec.RollbackLimit != nil &&
+		cr.Status.Synced && cr.Status.AtProvider.State == release.StatusFailed {
+		e.logger.Debug("Last release failed")
+		if cr.Status.Failed < *cr.Spec.RollbackLimit {
```
ktintc, line 255: here we could reuse rollbackLimitReached func mentioned above

apis/v1alpha1/types.go
turkenh: I agree that it may sound like a bool but I did follow the same convention as Kubernetes Jobs: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/types.go#L185
While checking yaml, it will be displayed as 1,2,3,4..., so may be it is not a big deal. 

Considering above, do you think we should still change? I am not too opinionated just tend to follow existing conventions.

examples/sample/release.yaml
turkenh: No, this was intentional since this is an example file.

pkg/controller/observe.go
turkenh: Good point, could remove 3 of 4 by increasing 10 --> 15

pkg/controller/observe.go
turkenh: 👍 Done

pkg/controller/release.go
```diff
@@ -240,8 +249,26 @@ func (e *helmExternal) Update(ctx context.Context, mg resource.Managed) (managed
...
+			// Rollback
+			e.logger.Debug("Will rollback to retry")
+			cr.Status.Failed++
+			if cr.Status.AtProvider.Revision < 2 {
```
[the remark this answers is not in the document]
turkenh: If revision is 1 (or less than 2 in the original impl :) ), rollback wouldn't work since there is no previous revision to rollback to. So, we need to check and uninstall in this case and use rollback otherwise.

pkg/controller/release.go
turkenh: `chartAction` sounds like we are doing something on charts (e.g. pull, pack, unpack) and but it is in fact a deployment action. 
So, I am fine with action too, renamed as `deployAction`.

PR #26, opened 2020-09-10
Remove local dev commands that do not use build sumodule

hasheddan:
Now that the build submodule handles starting a local dev env we should
remove the commands that provide similar functionality without it.

These also look like they were using `provider-gcp` anyway :)

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>


turkenh (approved):
Thank you!

PR #30, opened 2020-09-26
Match crossplane/crossplane build and GitHub configuration

negz:
Updates PR and issue templates as well as GitHub workflows and build submodule to match crossplane/crossplane#1775.


hasheddan (approved):

negz:
I've also updated https://codecov.io/account/gh/crossplane-contrib/yaml to match https://codecov.io/account/gh/crossplane/yaml - i.e. mostly to stop Codecov replying with comments on PRs.

PR #53, opened 2020-10-23
Concurrent processing of helm releases

### LINKED ISSUE #2: Parallel processing of multiple helm releases
Processing a Helm `Release` resource could take some time (pulling chart, installing/upgrading helm release). Since they are independent units, we should be able to process them concurrently without waiting one to complete.

[`MaxConcurrentReconciles` option of controller runtime ](https://github.com/kubernetes-sigs/controller-runtime/blob/56938944ad73232915fb2dce719812f99cbf727a/pkg/controller/controller.go#L36)seems like the solution here which we should configure. We can set a sensible default (e.g. 10) or expose it as a configuration option.

turkenh:

Fixes #2

I have:

### DIFF
--- a/pkg/controller/release/release.go
+++ b/pkg/controller/release/release.go
@@ -18,6 +18,7 @@ package release
 
 import (
 	"context"
+	"time"
 
 	"github.com/pkg/errors"
 	"helm.sh/helm/v3/pkg/chart"
@@ -43,7 +44,12 @@ import (
 	helmClient "github.com/crossplane-contrib/provider-helm/pkg/clients/helm"
 )
 
-const maxConcurrency = 10
+const (
+	maxConcurrency = 10
+
+	resyncPeriod     = 10 * time.Minute
+	reconcileTimeout = 2 * time.Minute
+)
 
 const (
 	errNotRelease                 = "managed resource is not a Release custom resource"
@@ -85,6 +91,8 @@ func Setup(mgr ctrl.Manager, l logging.Logger) error {
 			newHelmClientFn: helmClient.NewClient,
 		}),
 		managed.WithLogger(logger),
+		managed.WithTimeout(reconcileTimeout),
+		managed.WithLongWait(resyncPeriod),
 		managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))))
 
 	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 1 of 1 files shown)


pkg/controller/release/release.go
```diff
@@ -42,6 +44,13 @@ import (
 	helmClient "github.com/crossplane-contrib/provider-helm/pkg/clients/helm"
 )
 
+const (
+	maxConcurrency = 10
```
hasheddan, line 48: Just curious, what made you go with 10 here?

pkg/controller/release/release.go
```diff
@@ -42,6 +44,13 @@ import (
...
+
+	resyncPeriod     = 10 * time.Minute
```
hasheddan, line 50: Do you think this frequent of a resync period is needed? I'm guessing controllers already requeue appropriately when they need to periodically check

pkg/controller/release/release.go
turkenh: Nothing special, I just some number bigger than 1 (e.g. concurrent), not to block other release's because of a 
(or some) long running one (e.g. with a pre-install hook). 
I believe we may end up having this configurable somehow but for now, just trying to set some reasonable default.

pkg/controller/release/release.go
turkenh: It is a little hard to scale up with current default (1 minute, `longWait` in managed reconciler) since when the number of `Release` resources increases (imaging 1000 release's), controller will have them back into its queue before processing all and end up being blocked with infinite processing loop. 

Similarly, I believe this should also be a configurable option depending on the use case and trying to find some sensible defaults until we have ways to configure.

hasheddan: Hmm is setting a more frequent resync going to address this issue? It seems like it could actually make it worse https://github.com/kubernetes-sigs/controller-runtime/issues/521

turkenh: I am actually setting managed reconciler `longWait` with [this](https://github.com/crossplane-contrib/provider-helm/pull/53/files#diff-7447a2eaf6d31130caf30f51cecffc5770f6538354ddae88487c5698d477357dR95). I named it `resyncPeriod` as I believe they doing the same thing. 
So I am trying to increase from 1 min to 10 min.

hasheddan: Oh 😅 did not realize what file I was looking at and just assumed this was for the controller manager! Thanks for clarifying :)

hasheddan (approved):

PR #54, opened 2020-10-24
Update CI and Release Actions

negz:

This PR should improve some build times, and fix the promote action. It also allows us more flexibility in the tag names we create - i.e. the action can now be used to create RC tags before cutting a release branch.

I have:

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -6,10 +6,9 @@ on:
       version:
         description: 'Release version (e.g. v0.1.0)'
         required: true
-
-env:
-  UPBOUND_BOT_USR: ${{ secrets.UPBOUND_BOT_USR }}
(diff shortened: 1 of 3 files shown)


.github/workflows/tag.yml
```diff
@@ -18,21 +17,10 @@ jobs:
...
-          git config user.name "upbound-bot"
-          git config user.email "info@crossplane.io"
 
       - name: Create Tag
-        run: make -j2 tag
-        env:
-          VERSION: ${{ github.event.inputs.version }}
-        
\ No newline at end of file
+        uses: negz/create-tag@v1
```
hasheddan, line 22: Woo! Does this create annotated tags?

negz: https://github.com/negz/create-tag/blob/main/src/main.ts#L25

I believe so.

hasheddan: Nice!

hasheddan (approved):

PR #60, opened 2020-11-03
[release-0.3] Increase reconcile timeout to 10m

turkenh:
Signed-off-by: Hasan Turken <turkenh@gmail.com>
(cherry picked from commit a7810914a2fd3e3e94801149a3bf5cc981cbbaa5)

### DIFF
--- a/pkg/controller/release/release.go
+++ b/pkg/controller/release/release.go
@@ -18,6 +18,7 @@ package release
 
 import (
 	"context"
+	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
@turkenh now that you can configure a `Provider` at install time, it may make sense to make the value configurable via flag or env var in the future (https://github.com/crossplane/crossplane/blob/61109ab8bedc40fc8c6c7b3c861b20264c950177/apis/pkg/v1alpha1/config.go#L27)

Issue #63, opened 2020-11-04
Asynchronous deployment of helm charts


turkenh:
### What problem are you facing?
Currently, install and upgrade calls using helm go client is blocking which means, reconcile loop is being blocked until deployment is done. Since we don't set `wait` ([related](https://github.com/crossplane-contrib/provider-helm/issues/4)), this is fine if chart does not contain hooks. However, with hooks we encounter problems like [this](https://github.com/crossplane-contrib/provider-helm/issues/58#issuecomment-721083747).

### How could Crossplane help solve your problem?
We need to make this action asynchronous somehow.

turkenh:
Related: https://github.com/helm/helm/issues/5891

jbw976:
Are you still working on this @turkenh, or would it be available for someone else to take on?

turkenh:
For posterity, I believe this could be implemented by following a similar approach used in Terrajet: https://github.com/crossplane/terrajet/blob/6142eefbcf5b4e6490a4a39f8ceffa8810911d5f/pkg/terraform/workspace.go#L106

We might even consider to have a cli execution library (sync/async) as a reusable piece of code, e.g. in crossplane-runtime. Not sure if this worths the amount of work required though.

PR #64, opened 2020-11-05
Add wait option

### LINKED ISSUE #4: Helm wait option

turkenh:

Fixes #4

I have:

Locally with sample wordpress release.


srueg:
Nice one! 👍 

Any chance we can add an optional `timeout` override together with this? If `wait` is enabled the default of `5m0s` might be too short.

turkenh:
> Any chance we can add an optional timeout override together with this? If wait is enabled the default of 5m0s might be too short.

Hmm, yes we can. 
But here comes the challenge, timeout should not be longer than [reconcileTimeout](https://github.com/crossplane-contrib/provider-helm/blob/master/pkg/controller/release/release.go#L51) (which is 10m right now), otherwise, deployment will be cancelled afterwards. We can not adapt it per resource config, e.g. if timeout > reconcileTimeout - use timeout, because It is not possible to configure it per resource.

We could increase reconcile timeout to a reasonable max (e.g. `30m`) and document wait timeout cannot be more than that. However, I am not happy with this and may be we should fix #63 first. 

Here is one concrete problematic scenario (which is already there if there are hooks):
1. Create a helm release with wrong image tag 
    i. Reconciler waits deployment succeeds until timeout (let's say configured as 10m)
    ii. At the same time, you realized the problem and updated `Release` values with correct tag
2. Your new update will only be processed after timeout which would be quite confusing and a bad UX.

Ideally, new update should be able to cancel previous deployment but this should be decided carefully, i.e. only when there is a configuration drift etc.

Anyway, not all is about making timeouts configurable, but just wanted to share why we need a better solution here :)

srueg:
I see the problem. And I agree that a proper solution with #63 makes more sense.

srueg (approved):

lukeweber (approved):

PR #67, opened 2020-11-10
Incluster provisioning with InjectedIdentity

### LINKED ISSUE #62: In cluster provisioning with RBAC

It is very possible to use provider-helm for in-cluster provisioning as a regular helm controller instead of deploying into external cluster. Today, this could be realized by putting kubeconfig of the local cluster into a secret and creating a `ProviderConfig` pointing to that secret. However, this approach is not the expected way to communicate/authenticate cluster api from [inside the cluster](https://github.com/kubernetes/client-go/tree/master/examples/in-cluster-client-configuration).

Is should be possible to do in cluster provisioning by relying on RBAC. 
This could be achievable by introducing a new `credentials.source`, e.g. "InCluster" and configuring provider controller with proper service account during deployment.

turkenh:

Fixes #62 

I have:

Todo: 

- [x] Update doc


hasheddan (approved):
LGTM! Nice work @turkenh :)

PR #69, opened 2020-11-24
Preserve unknown fields in Helm values

srueg:
Preserve unknown fields in Helm values.
Otherwise the values will be dropped by the API as all its fields are unknown (no schema). This behavior changed with the v1 CRD introduced in #65 .

See also: https://github.com/crossplane/crossplane/blob/master/apis/apiextensions/v1beta1/composition_types.go#L85

I have:

Manually on a cluster.

### DIFF
--- a/apis/release/v1alpha1/types.go
+++ b/apis/release/v1alpha1/types.go
@@ -60,6 +60,7 @@ type SetVal struct {
 
 // ValuesSpec defines the Helm value overrides spec for a Release
 type ValuesSpec struct {
+	// +kubebuilder:pruning:PreserveUnknownFields
 	Values     runtime.RawExtension `json:"values,omitempty"`
 	ValuesFrom []ValueFromSource    `json:"valuesFrom,omitempty"`
 	Set        []SetVal             `json:"set,omitempty"`
--- a/package/crds/helm.crossplane.io_releases.yaml
+++ b/package/crds/helm.crossplane.io_releases.yaml
@@ -181,6 +181,7 @@ spec:
                     type: array
                   values:
                     type: object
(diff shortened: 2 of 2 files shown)


apis/release/v1alpha1/types.go
```diff
@@ -60,6 +60,7 @@ type SetVal struct {
 
 // ValuesSpec defines the Helm value overrides spec for a Release
 type ValuesSpec struct {
+	// +kubebuilder:pruning:PreserveUnknownFields
```
hasheddan, line 63: Should this be on the overall `Release` struct with the other `kubebuilder` annotations?

srueg: As I understood it, with v1 CRDs this is a field level configuration.

srueg: From the docs:
> By default, all unspecified fields for a custom resource, across all versions, are pruned. It is possible though to opt-out of that for specifc sub-trees of fields by adding x-kubernetes-preserve-unknown-fields: true in the structural OpenAPI v3 validation schema.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning

hasheddan: ah nice, looks like you are correct :+1:

hasheddan (approved):

negz (approved):
Thanks for catching this @srueg!

apis/release/v1alpha1/types.go
hasheddan: We actually hit this same issue in c/c but it was caught before release https://github.com/crossplane/crossplane/pull/1858

PR #73, opened 2020-12-11
Create connection secret with desired fields

### LINKED ISSUE #56: Create Connection Secrets

Currently there's no way to create a connection secret from a resource provisioned by provider-helm. For example using the [MariaDB Chart](https://github.com/bitnami/charts/tree/master/bitnami/mariadb) I'd like to expose the username, password and hostname in a connection secret.

Reading state back from created resources by a Helm chart provider-helm could create a connection secret.
Might be related to https://github.com/crossplane/crossplane/issues/1722 and/or https://github.com/crossplane/crossplane/issues/1603.

turkenh:
Signed-off-by: Hasan Turken <turkenh@gmail.com>

Closes #56

I have:


examples/sample/release.yaml
```diff
@@ -31,5 +31,28 @@ spec:
 #         name: svals
 #         namespace: wordpress
 #         optional: false
+#  connectionDetails:
```
lukeweber, line 34: I feel connectionDetails isn't the perfect name as it's really general purpose data sync from a Release.

releaseObservation:
 ...
  writeObservationSecretToRef:

pkg/controller/release/observe.go
```diff
@@ -31,6 +40,7 @@ const (
 	errReleaseInfoNilInObservedRelease = "release info is nil in observed helm release"
 	errChartNilInObservedRelease       = "chart field is nil in observed helm release"
 	errChartMetaNilInObservedRelease   = "chart metadata field is nil in observed helm release"
+	errObjectNotPartOfRelease          = "object is not part of release: %v"
```
lukeweber, line 43: Really happy with where this landed for security reasons and minimizing unexpected behavior. 

I could imagine someone might install a few helm charts and create a final observation, but this puts some rails that no one mis-configures and fetches something from a different helm chart.

The downside of course as discussed is that maybe an operator is installed and simply creates an object that isn't part of the release explicitly. If this ends up being a use case, they could provide an empty secret in the chart, or we could consider in the future an escape config to let them fetch arbitrary resources, but this seems like a great first step.

examples/sample/release.yaml
turkenh: I see what you mean and actually give it a try to change it in that direction, however ended up with current solution because of following reasons:

- `writeConnectionSecretToRef` is part of [crossplane managed resource spec](https://github.com/crossplane/crossplane-runtime/blob/master/apis/common/v1/resource.go#L137), there are existing machinery coming with crossplane runtime hence this field is common across all providers. I preferred to reuse this pattern/machinery here.
- `connectionDetails` is also something crossplane users should be familiar with, it exists in composite resource api ([example](https://github.com/upbound/platform-ref-aws/blob/master/cluster/composition.yaml#L14)) with similar purposes, where you select what you want to be included in the connection secret that composite resource creates.

hasheddan (commented):
@turkenh changing this to `v1beta1` should cause integration tests to start passing :)

https://github.com/crossplane-contrib/provider-helm/blob/5df147edcf07f78de8ed37b648e8ce78a54c6670/cluster/integration/integration_tests.sh#L274

lukeweber (approved):

pkg/controller/release/observe.go
emilkje: I'm interested in the scenario where I want to observe and provision connection details from a Secret created by an operator. In this case I want to retrieve the kubeconfig from a [vcluster](https://github.com/loft-sh/vcluster/tree/main/charts/k3s) Release. I'm not confident that [the controller](https://github.com/loft-sh/vcluster/blob/3121810c78b63486213d54aa98472ef237d447da/pkg/util/kubeconfig/kubeconfig.go#L27) could handle an existing secret without a patch either. Do I have a way to prevent `object is not part of release` error without forking the charts and creating a dummy secret? It seems like an inelegant and operationally expensive solution.

examples/sample/release.yaml
```diff
@@ -31,5 +31,28 @@ spec:
...
+#      name: wordpress-example
+#      namespace: wordpress
+#      fieldPath: spec.ports[0].port
+#      toConnectionSecretKey: port
+#    - apiVersion: v1
+#      kind: Secret
+#      name: wordpress-example
+#      namespace: wordpress
+#      fieldPath: data.wordpress-password
+#      toConnectionSecretKey: password
```
jonnylangefeld, line 53: I have a `ServiceAccountKey.cloudplatform.gcp.upbound.io/v1beta1` inside a helm chart, which creates a connection secret at `<uid>-serviceaccountkey` in the `crossplane` namespace.

So I made a list entry like this on my `Release`:
```
            - apiVersion: v1
              kind: Secret
              name: ""
              namespace: crossplane
```
And then I'm patching the name like so:
```
        - type: CombineFromComposite
          combine:
            variables:
              - fromFieldPath: metadata.uid
            strategy: string
            string:
              fmt: "%s-serviceaccountkey"
          toFieldPath: spec.connectionDetails[0].name
          policy:
            fromFieldPath: Required
```

Now the issue is that on the `Release` I get an event that says:
```
│   Warning  CannotObserveExternalResource  31s (x5 over 3m34s)    managed/release.helm.crossplane.io  cannot get connection details: object is not part of release: {Secret crossplane 8578b338-7520-44e8-a3e7-a1555abfd257-serviceaccountkey  v1  }
```
Which is of course true, because the connection secret is not part of that release, it's just a result of it.
But the connection secret exists:
```
╰─ kg secret -n crossplane 8578b338-7520-44e8-a3e7-a1555abfd257-serviceaccountkey
NAME                                                     TYPE                                DATA   AGE
8578b338-7520-44e8-a3e7-a1555abfd257-serviceaccountkey   connection.crossplane.io/v1alpha1   2      7m33s
```

How can I propagate the secret of that `Secret` resource outside of the release? I thought it should be possible since I'm just referring to an arbitrary secret by name and namespace that the controller has access to...

turkenh: As we discussed offline, `skipPartOfReleaseCheck: true` should do the trick.

PR #79, opened 2021-02-07
Support URL in chart spec

### LINKED ISSUE #78: Installing Helm release from URL causes upgrade error
### What happened?
When creating a new Release with a URL in the `name` field, I receive a `no such file or directory` error message due to assumptions about release filenames.

### How can we reproduce it?
Install a helm chart from a URL:

```yaml
 kind: Release
 metadata:
   name: test
   namespace: default
 spec:
   providerConfigRef:
     name: helm-provider
   forProvider:
     namespace: default
     chart:
       name: https://gitlab...../api/v4/projects/..../packages/generic/helm-chart/0.25.0/chart-bleeding.tgz?private_token=....
       repository: ""
       version: ""
```

Produces following log messages on Release:

```
  Normal   CreatedExternalResource       7s               managed/release.helm.crossplane.io  Successfully requested creation of external resource
  Normal   UpdatedExternalResource       5s               managed/release.helm.crossplane.io  Successfully requested update of external resource
  Warning  CannotUpdateExternalResource  4s (x3 over 7s)  managed/release.helm.crossplane.io  failed to upgrade release: failed to load chart: stat /tmp/charts/https:/gitlab....?private_token=....-0.25.0.tgz: no such file or directory
```

Note that the chart runs successfully first time and creates the relevant resources. It looks like this may be caused by late init modifying the `version` field from the empty string, causing different behaviour.

### What environment did it happen in?
Crossplane / Provider-helm version: master 

### Issues
2 Problems here - first is that repository and version are not optional in the `Release` API, but they _are_ optional when it comes to calling Helm, because Helm supports installing a chart `.tar.gz` directly from URL. Setting them both to empty strings sorta works but...

Second is that the `chartFilePath` logic in `PullAndLoadChart` (Helm client) uses the name field incorrectly if a version is set. 

The version is late-inited from the chart itself so it is not possible to avoid this error by setting an empty version string.

It also makes an assumption about the filename format (`${name}-${version}.tgz`) of the Helm chart in a repository which doesn't necessarily hold true - AIUI, a Helm repository can set whatever filename it likes for a chart version entry, and I think the name in the repository is what is used when [downloading](https://github.com/helm/helm/blob/46d80f68c11d8107b688b9c8299fa02bccba312f/pkg/downloader/chart_downloader.go#L109).

### Fix?
 * `repository` and `version` should be optional (string pointer, `omitempty`, `// +optional`)
 * Chart filename logic needs expanding, taking into account that name can contain a URL, an absolute path or a path within a repository.

Happy to look at PR-ing a fix for these but I don't have a clear view on what the 'Chart name logic' behaviour should be. I suspect it needs to support something similar to [ResolveChartVersion](https://github.com/helm/helm/blob/46d80f68c11d8107b688b9c8299fa02bccba312f/pkg/downloader/chart_downloader.go#L156) in Helm itself.

turkenh:

Fixes #78

I have:

### DIFF
--- a/apis/release/v1beta1/types.go
+++ b/apis/release/v1beta1/types.go
@@ -27,9 +27,15 @@ import (
 
 // A ChartSpec defines the chart spec for a Release
 type ChartSpec struct {
-	Repository    string               `json:"repository"`
-	Name          string               `json:"name"`
-	Version       string               `json:"version"`
+	// Repository: Helm repository URL, required if ChartSpec.URL not set
+	Repository string `json:"repository,omitempty"`
+	// Name of Helm chart, required if ChartSpec.URL not set
+	Name string `json:"name,omitempty"`
+	// Version of Helm chart, late initialized with latest version if not set
+	Version string `json:"version,omitempty"`
+	// URL to chart package (typically .tgz), optional and overrides others fields in the spec
+	URL string `json:"url,omitempty"`
+	// PullSecretRef is reference to the secret containing credentials to helm repository
 	PullSecretRef xpv1.SecretReference `json:"pullSecretRef,omitempty"`
 }
 
--- a/examples/sample/release.yaml
+++ b/examples/sample/release.yaml
@@ -12,6 +12,7 @@ spec:
 #     pullSecretRef:
 #       name: museum-creds
 #       namespace: default
+#     url: "https://charts.bitnami.com/bitnami/wordpress-9.3.19.tgz"
     namespace: wordpress
 #    wait: true
     values:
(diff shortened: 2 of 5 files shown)


apis/release/v1beta1/types.go
```diff
@@ -27,9 +27,10 @@ import (
 
 // A ChartSpec defines the chart spec for a Release
 type ChartSpec struct {
-	Repository    string               `json:"repository"`
-	Name          string               `json:"name"`
-	Version       string               `json:"version"`
+	Repository    string               `json:"repository,omitempty"`
+	Name          string               `json:"name,omitempty"`
+	Version       string               `json:"version,omitempty"`
+	URL           string               `json:"url,omitempty"`
```
hasheddan, line 33: Might be worth adding comments on these to indicate when fields are required (e.g. if not using URL you need to provide Repository / Name / Version).

hasheddan (approved):
One nitpick but it is only for documentation, LGTM!

PR #86, opened 2021-04-16
Label created namespaces

### LINKED ISSUE #77: Mark Created Namespaces

If a chart is uninstalled, the previously created namespace won't be deleted with it. This might be a good thing because other resources like PVCs are also left in place (if they're created by a StatefulSet for example).
Nevertheless we want to remove these orphaned namespaces at some point.

If a release was deleted and the chart was successfully uninstalled, the provider could add a label & annotation to the namespace:
```yaml
metadata:
  labels:
    meta.helm.sh/deleted: 'true'
  annotations:
    meta.helm.sh/deletionTimestamp: "2021-01-18T13:25:35Z"
```

This would help to manually or automatically remove namespaces of uninstalled charts:
```
kubectl delete ns -l meta.helm.sh/deleted=true
```

### LINKED ISSUE #80: Optional Namespace Creation
provider-helm tries to create a namespace even if it already exists. In general this is fine because it fails gracefully and continues with the rest of the execution [1], but it still requires namespace creation RBAC permissions.
In some cases I only want to deploy to existing namespaces and don't want to grant this permission to provider-helm.

Currently namespace creation is hardcoded here: https://github.com/crossplane-contrib/provider-helm/blob/94c5b9a8406dd5885fe4e2a43a997fd0960af56a/pkg/clients/helm/client.go#L103

I see two options how to solve this:
1. Introduce an option to configure if a release should try to create the namespace
2. Let the provider figure out if the namespace creation is actually necessary beforehand

[1] https://github.com/helm/helm/blob/master/pkg/action/install.go#L307

srueg:
Add a label to created namespaces.

Fixes #77
Fixes #80

I have:

### DIFF
--- a/examples/sample/release.yaml
+++ b/examples/sample/release.yaml
@@ -14,7 +14,8 @@ spec:
 #       namespace: default
 #     url: "https://charts.bitnami.com/bitnami/wordpress-9.3.19.tgz"
     namespace: wordpress
-#    wait: true
+#   skipCreateNamespace: true
+#   wait: true
     values:
       service:
         type: ClusterIP
--- a/apis/release/v1beta1/types.go
+++ b/apis/release/v1beta1/types.go
@@ -75,11 +75,17 @@ type ValuesSpec struct {
 
 // ReleaseParameters are the configurable fields of a Release.
 type ReleaseParameters struct {
-	Chart       ChartSpec         `json:"chart"`
-	Namespace   string            `json:"namespace"`
-	Wait        bool              `json:"wait,omitempty"`
+	Chart ChartSpec `json:"chart"`
+	// Namespace to install the release into.
+	Namespace string `json:"namespace"`
+	// SkipCreateNamespace won't create the namespace for the release. This requires the namespace to already exist.
+	SkipCreateNamespace bool `json:"skipCreateNamespace,omitempty"`
+	// Wait for the release to become ready.
+	Wait bool `json:"wait,omitempty"`
+	// PatchesFrom describe patches to be applied to the rendered manifests.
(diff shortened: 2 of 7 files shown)


turkenh:
@srueg thanks!
Would you also mind tackling #80 by introducing a new field to spec? 
I believe it could be something like `spec.dontCreateNamespace` which is, by default false, and results in creation of namespace.

srueg:
> @srueg thanks!
> Would you also mind tackling #80 by introducing a new field to spec?
> I believe it could be something like `spec.dontCreateNamespace` which is, by default false, and results in creation of namespace.

Sure thing 👍 
I'd vote for `spec.skipCreateNamespace` though?

turkenh:
> I'd vote for `spec.skipCreateNamespace` though?

Sounds good 👍
Thank you!

examples/sample/release.yaml
```diff
@@ -14,7 +14,8 @@ spec:
 #       namespace: default
 #     url: "https://charts.bitnami.com/bitnami/wordpress-9.3.19.tgz"
     namespace: wordpress
-#    wait: true
+#   skipCreateNamespace: true
+    wait: true
```
turkenh, line 18: I'd prefer to keep wait disabled in the example, if there is no special reason.

turkenh (approved):
@srueg LGTM, thanks for working on this 👍

examples/sample/release.yaml
turkenh: /ping @srueg

srueg: The reason I added it was so that the e2e test actually waits for the release to be installed. 
But I agree, it probably doesn't make for a good example.

turkenh:
Just released v0.7.0 with this.

PR #87, opened 2021-05-12
Use recommended packaging format

negz:

Per https://github.com/crossplane/crossplane/pull/2299

I have:

I'm confident enough from testing a similar change at https://github.com/crossplane/provider-azure/pull/244.


package/crossplane.yaml
```diff
@@ -3,6 +3,32 @@ kind: Provider
...
+    friendly-name.meta.crossplane.io: Provider Helm
+    friendly-group-name.meta.crossplane.io/helm.crossplane.io: Helm Charts
+    friendly-kind-name.meta.crossplane.io/release.helm.crossplane.io: Helm Release
```
hasheddan, line 29: Does it make sense for `helm.crossplane.io` group to have a friendly name of `Helm Charts`? Maybe so, but it does feel a little strange for `Helm Release` to be a kind in the `Helm Chart` group.

negz: Okay, I removed "Charts".

hasheddan (approved):

PR #98, opened 2021-07-14
AuthProvider for GKE Clusters created by provider-gcp

AaronME:

This is an AuthProvider based on the [GCP Auth Provider](https://pkg.go.dev/k8s.io/client-go@v0.21.2/plugin/pkg/client/auth/gcp) which removes the dependency on Default credentials created by the call to [google.DefaultTokenSource](https://github.com/kubernetes/client-go/blob/7a90b08589bbf0d36a27041d5491b7b8d5430f15/plugin/pkg/client/auth/gcp/gcp.go#L165). 

This code introduces a new, optional field to the ProviderConfig called `google-app-creds`, which can take a reference to a secret key holding a client-credentials.json:

```
apiVersion: helm.crossplane.io/v1alpha1
kind: ProviderConfig
metadata:
  name: example
  namespace: upbound-system
spec:
  credentials:
    secretRef:
      key: kubeconfig
      name: 01c3f1bc-406b-401c-a44c-96eaa46f4bcc-gkecluster
      namespace: upbound-system
    source: Secret
  google-app-creds:
    source: Secret
    secretRef:
      key: creds
      name: test-service-account-creds
      namespace: upbound-system
  ```
  
  The contents of test-service-account-creds["creds"] looks like the following (base64-encoded):
```
{
  "type": "service_account",
  "project_id": "project-id-value",
  "private_key_id": "...",
  "private_key": "-----BEGIN PRIVATE KEY-----\n...\n-----END PRIVATE KEY-----\n",
  "client_email": "service-account-name@project-id-value.iam.gserviceaccount.com",
  "client_id": "123456789098765432101",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://oauth2.googleapis.com/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/..."
}
```

Fixes https://github.com/crossplane-contrib/provider-helm/issues/72

I have:

Deployed an app via provider-helm using the aboveProviderConfig and manually installed client-credentials.json secret.

### CODE REVIEW

File: apis/v1alpha1/types.go [hunk 1]
```diff
@@ -30,6 +30,7 @@ type ProviderConfigSpec struct {
 // A ProviderConfigStatus defines the status of a Provider.
 type ProviderConfigStatus struct {
 	xpv1.ProviderConfigStatus `json:",inline"`
+	GoogleAppCreds            xpv1.ProviderCredentials `json:"google-app-creds,omitempty"`
```
negz, line 33: Is this intended to be in status (as well as spec)?

apis/v1beta1/types.go
```diff
@@ -25,6 +25,7 @@ import (
 // A ProviderConfigSpec defines the desired state of a Provider.
 type ProviderConfigSpec struct {
 	xpv1.ProviderConfigSpec `json:",inline"`
+	GoogleAppCreds          xpv1.ProviderCredentials `json:"google-app-creds,omitempty"`
```
negz, line 28: Would it make sense for this to be a new kind of `source` within the existing `credentials` field? Something like `source: GoogleAppSecret`? This would require duplicating rather than embedding most of the `xpv1.ProviderConfigSpec`, but doing so is totally fine.

pkg/controller/release/auth_provider.go
```diff
@@ -0,0 +1,245 @@
...
+	"k8s.io/klog/v2"
+)
+
+const (
+	providerGCPAuthPlugin = "provider-gcp" // so that this is different than "gcp" that's already in client-go tree.
+)
+
+func init() {
+	if err := restclient.RegisterAuthProviderPlugin(providerGCPAuthPlugin, newProviderGCPAuthProvider); err != nil {
+		klog.Fatalf("Failed to register gcp auth plugin: %v", err)
```
negz, line 39: Is using klog mandatory in this context or could we use the logging package from crossplane-runtime?

pkg/controller/release/auth_provider.go
```diff
@@ -0,0 +1,245 @@
...
+	return &providerGCPAuthProvider{cts, persister}, nil
+}
+
+func tokenSource(providerGCPConfig map[string]string) (oauth2.TokenSource, error) {
+	// Google Application Credentials-based token source
+	scopes := parseScopes(providerGCPConfig)
+	ts, err := google.JWTAccessTokenSourceWithScope([]byte(providerGCPConfig["creds"]), scopes...)
+	if err != nil {
+		return nil, fmt.Errorf("cannot construct JWT token source: %v", err)
+	}
```
negz, line 108: Can we use `errors.Wrapf` instead of `fmt.Errorf`? We mostly use `github.com/pkg/errors` throughout the Crossplane codebases.

negz (commented):
This is awesome, thank you @aaronme! I haven't done a thorough review since this is still a draft; just a few comments based on a quick skim.

turkenh:
Thank you @AaronME for tackling this!

We will need the same functionality in provider-kubernetes or other possible k8s based providers. So, I am wondering your thoughts on making this part of provider-gcp or crossplane-runtime.

apis/v1beta1/types.go
AaronME: I've created https://github.com/crossplane/crossplane-runtime/issues/281 to look at a new CredentialSource specific to Google Cloud.

AaronME:
> Thank you @AaronME for tackling this!
> 
> We will need the same functionality in provider-kubernetes or other possible k8s based providers. So, I am wondering your thoughts on making this part of provider-gcp or crossplane-runtime.

@turkenh , I think that's a good idea. Opened an issue on crossplane-runtime: https://github.com/crossplane/crossplane-runtime/issues/215

PR #109, opened 2021-09-14
Support GKE OAuth Authorization

### LINKED ISSUE #72: Support GKE OAuth Authorization

https://cloud.google.com/kubernetes-engine/docs/how-to/api-server-authentication

As far as I can tell the only way to get a GKE cluster to produce a kubeconfig that can be used by provider-helm (i.e. a kubeconfig that authenticates as cluster-admin) is to enable basic master auth, support for which was removed in v1.19. GKE can generate a client certificate instead, but that is also deprecated and doesn't actually have permissions to do anything useful per https://github.com/kubernetes/kubernetes/issues/65400.

I believe this means that provider-helm can't be used to release charts to a GKE 1.19 or above cluster created via provider-gcp.

I believe it may be possible to integrate with GKE's native OAuth support somehow? Crossplane can create a GCP `ServiceAccount` - perhaps we could have that write its OAuth token to a connection secret that this provider could consume? I believe based on very cursory research that that token can be used to authenticate to the API server, and that the resulting user will be granted RBAC access corresponding to their GCP IAM permissions.

negz:

Fixes #72
Closes #98

This commit adds an optional set of 'identity' credentials to the Helm provider config. Currently only the GoogleApplicationCredentials type of identity is supported. If supplied, the kubeconfig (fetched from the regular 'credentials') is used to create an API Server REST client, which is wrapped such that it will use a Google Application Credentials file to fetch OAuth tokens which are then used as the Bearer token to authorize requests.

@AaronME opened #98 to fix this, but we decided per https://github.com/crossplane/crossplane-runtime/issues/281 that it made sense to add support to crossplane-runtime instead so that we could reuse code across other Kubernetes-ey providers, like provider-kubernetes. After messing with this all day I've come full circle and decided it's better to just bake this straight into each provider for now. I'll add a comment in https://github.com/crossplane/crossplane-runtime/issues/281 explaining why.

My implementation differs from #98 mostly because per https://github.com/crossplane/crossplane-runtime/issues/281#issuecomment-918645907 I noticed the authentication provider method was deprecated, and while trying to figure out why it was deprecated and what the alternatives were I stumbled upon this seemingly much simpler approach.

I have:

I've used the updated example `ProviderConfig` to deploy the example `Release` to a GKE cluster.


apis/v1alpha1/types.go
```diff
@@ -24,8 +24,15 @@ import (
...
-	// Credentials required to authenticate to this provider.
+	// Credentials used to connect to the Kubernetes API. Typically a
+	// kubeconfig file. Use InjectedIdentity for in-cluster config.
 	Credentials ProviderCredentials `json:"credentials"`
+
+	// Identity used to authenticate to the Kubernetes API. The identity
+	// credentials can be used to supplement kubeconfig 'credentials', for
+	// example by configuring a bearer token source such as OAuth.
+	// +optional
+	Identity *Identity `json:"identity,omitempty"`
```
negz, line 35: Currently these changes make a ProviderConfig look like this:


```yaml
kind: ProviderConfig
metadata:
  name: default
spec:
  credentials:
    source: Secret
    secretRef:
      name: cluster-credentials
      namespace: crossplane-system
      key: kubeconfig
  identity:
    type: GoogleApplicationCredentials
    source: Secret
    secretRef:
      name: gcp-credentials
      namespace: crossplane-system
      key: credentials.json
```

This is probably the safest and most backward compatible API, but I don't like it for a few reasons. Namely:

1. `credentials` and `identity` sound very similar. It's not obvious what is for what.
2. `credentials` is a misnomer for a kubeconfig, which configures more than creds, and may configure no creds.
3. I could imagine wanting to configure this provider (and others like provider-kubernetes) by just providing an API server URL etc without a full kubeconfig.

I think we could address my concerns above by making a larger change to the ProviderConfig API. This change would still be backward compatible, but in a kind of awkward way. It would instead look something like this:

```yaml
kind: ProviderConfig
metadata:
  name: default
spec:
  # 'kubernetes' is a new optional field.
  kubernetes:
    # Nesting kubeconfig here gives us the opportunity to specify Kubernetes configuration
    # differently in future - e.g. by loading an API server URL, CA cert, and token from different
    # connection detail keys.
    kubeconfig:
      # We support loading a kubeconfig from a Secret, env var or filesystem like we do for creds.
      source: Secret
      secretRef:
        namespace: crossplane-system
        name: cluster-creds
        key: credentials.json
  # 'credentials' becomes an optional field.
  # You must specify at least one of 'kubernetes' and 'credentials'.
  credentials:
    # The default type of credentials is Kubeconfig for backward compatibility, but we recommend
    # that folks use the 'kubernetes' field instead of Kubeconfig "credentials" type. The
    # 'credentials' field then becomes mostly a way to override or extend any credentials loaded
    # from a kubeconfig.
    type: GoogleApplicationCredentials
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: cluster-creds
      key: credentials.json
```

That all said, I still don't love the above API because it seems like 'credentials' should be under 'kubernetes'.

pkg/clients/client.go
```diff
@@ -53,11 +51,14 @@ func restConfigFromAPIConfig(c *api.Config) (*rest.Config, error) {
...
-		return nil, errors.New(fmt.Sprintf("cluster for currentContext (%s) not found", c.CurrentContext))
+		return nil, errors.Errorf("cluster for currentContext (%s) not found", c.CurrentContext)
 	}
 	user := c.AuthInfos[ctx.AuthInfo]
 	if user == nil {
-		return nil, errors.New(fmt.Sprintf("auth info for currentContext (%s) not found", c.CurrentContext))
+		// We don't require a user because it's possible user
+		// authorization configuration will be loaded from a separate
+		// set of identity credentials (e.g. Google Application Creds).
+		user = &api.AuthInfo{}
```
negz, line 61: This change might not actually be necessary. I interpreted https://github.com/crossplane/provider-gcp/issues/343 as saying that there would be no entry in the `users` list in some cases, but for our GKE cluster implementation it's more that the user has an empty `user` block - i.e.:

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: REDACTED
    server: REDACTED
  name: scale-test-0-c2rvh
contexts:
- context:
    cluster: scale-test-0-c2rvh
    user: scale-test-0-c2rvh
  name: scale-test-0-c2rvh
current-context: scale-test-0-c2rvh
kind: Config
preferences: {}
users:
- name: scale-test-0-c2rvh
  user: {}
```

turkenh (approved):
Looking good to me, also tested with both 1.18 and 1.20 GKE cluster and verified both are working.

Thanks a lot for tackling this @negz (and @AaronME for originally kicking this off)!

apis/v1alpha1/types.go
negz: > Not sure whether there is a valid use case for this (don't remember such a request from community)

Right - I'm not aware of any requests from the community either so it's definitely hypothetical. My thinking is that I could conceive of a managed resource that represents a Kubernetes cluster that just returned keys for things like API server URL and CA cert data, rather than forming a whole kubeconfig. This is actually how some of our cluster managed resources used to work waaaay back around Crossplane v0.2 or v0.3 before we standardised them all on exposing a fully formed kubeconfig.

pkg/clients/client.go
negz: I'm going to leave this out of laziness since it can't hurt. 🤷

PR #114, opened 2021-12-01
Add option to skip CRDs in a Release

### LINKED ISSUE #107: Skip CRDs in Release

I am using a chart which does not use the full CRD specs for the project. I want the full ones but I am unable to skip what the chart is creating. Then the release will reconcile and override the full ones I applied. 

Add an option to the Release kind to skip crds.
Like this: `spec.forProvider.skipCRDs: true`
This is an option on Helm itself with this flag: `--skip-crds`

somaritane:
Signed-off-by: Timofey Ilinykh <ilinytim@gmail.com>

This PR adds the ability to skip the installation of CRDs for a Release.
It is configured by `spec.forProvider.skipCRDs` field (`=false` by default).

The effect is equal to Helm's [`--skip-crds`](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) option

Fixes #107 

I have:

Tested against local dev cluster using the example chart with CRD templates (examples/sample/release-with-crds.yaml).

### DIFF
--- a/pkg/clients/helm/client.go
+++ b/pkg/clients/helm/client.go
@@ -24,7 +24,6 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	"github.com/pkg/errors"
@@ -73,13 +72,22 @@ type client struct {
 	uninstallClient *action.Uninstall
 }
 
+// ArgsApplier defines helm client arguments helper
+type ArgsApplier func(*Args)
+
 // NewClient returns a new Helm Client with provided config
-func NewClient(log logging.Logger, config *rest.Config, namespace string, wait bool, timeout time.Duration) (Client, error) {
-	rg := newRESTClientGetter(config, namespace)
+func NewClient(log logging.Logger, restConfig *rest.Config, argAppliers ...ArgsApplier) (Client, error) {
+
+	args := &Args{}
+	for _, apply := range argAppliers {
+		apply(args)
+	}
+
+	rg := newRESTClientGetter(restConfig, args.Namespace)
 
 	actionConfig := new(action.Configuration)
 	// Always store helm state in the same cluster/namespace where chart is deployed
-	if err := actionConfig.Init(rg, namespace, helmDriverSecret, func(format string, v ...interface{}) {
+	if err := actionConfig.Init(rg, args.Namespace, helmDriverSecret, func(format string, v ...interface{}) {
 		log.Debug(fmt.Sprintf(format, v))
 	}); err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


somaritane:
Hi @turkenh, I might need your approval for the PR-related GH workflow run, as this is my first-time contribution, thank you!

somaritane:
@turkenh Looks like `go.lint` make target fails on timeout:
https://github.com/crossplane-contrib/provider-helm/runs/4378543937?check_suite_focus=true#step:11:18
> 08:15:42 [ .. ] golangci-lint
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
08:17:44 [FAIL]
make[2]: *** [go.lint] Error 1
build/makelib/golang.mk:192: recipe for target 'go.lint' failed

I can try to increase the timeout in [`.golangci.yaml`](https://github.com/crossplane-contrib/provider-helm/blob/master/.golangci.yml) via `run.timeout` option (I see `run.deadline = 2m` there, but it was [deprecated](https://github.com/golangci/golangci-lint/pull/793) long time ago)

turkenh:
> @turkenh Looks like `go.lint` make target fails on timeout for `e2e-tests` CI job: https://github.com/crossplane-contrib/provider-helm/runs/4378543937?check_suite_focus=true#step:11:18
> 
> > 08:15:42 [ .. ] golangci-lint
> > level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
> > 08:17:44 [FAIL]
> > make[2]: *** [go.lint] Error 1
> > build/makelib/golang.mk:192: recipe for target 'go.lint' failed
> 
> I can try to increase the timeout in [`.golangci.yaml`](https://github.com/crossplane-contrib/provider-helm/blob/master/.golangci.yml) via `run.timeout` option (I see `run.deadline = 2m` there, but it was [deprecated](https://github.com/golangci/golangci-lint/pull/793) long time ago)

Just retriggered and worked fine this time. 
Let's observe for a while before considering bumping timeout.

examples/sample/release-with-crds.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: argo-cd-example
+spec:
+  forProvider:
+    chart:
+      name: argo-cd
+      repository: https://charts.bitnami.com/bitnami
+      version: 2.0.12
+    namespace: argo-cd
+    skipCRDs: true
```
turkenh, line 12: Instead of new example yaml, please add the optional config as commented out [like this](https://github.com/crossplane-contrib/provider-helm/blob/e49c3e791f5c7bd95d50ab0762297daeb084c459/examples/sample/release.yaml#L17).

pkg/clients/helm/client.go
```diff
@@ -74,7 +74,7 @@ type client struct {
 }
 
 // NewClient returns a new Helm Client with provided config
-func NewClient(log logging.Logger, config *rest.Config, namespace string, wait bool, timeout time.Duration) (Client, error) {
+func NewClient(log logging.Logger, config *rest.Config, namespace string, wait bool, timeout time.Duration, skipCRDs bool) (Client, error) {
```
turkenh, line 77: Nit: 
With the number of parameters increasing in the function signature here, I believe we should switch using [variadic functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).

turkenh (approved):
Thanks a lot for your contribution @somaritane!

Looking good, just added two nits.

examples/sample/release-with-crds.yaml
somaritane: I was considering that option from the beginning. But the problem with `wordpress` chart is that it doesn't install CRDs. As result, `spec.forProvider.skipCRDs` won't have any effect on deployment, and the example won't be true to life.
That's why I've added a separate example with `argo-cd` chart. 
But sure, I can amend.

somaritane: Agree, amended. Using argo-cd helm as the original release example would require extra efforts on mapping all parameter examples to real-world argo-cd deployment cases, might be an extra enhancement outside of this PR :)

pkg/clients/helm/client.go
somaritane: @turkenh amended

PR #121, opened 2022-02-16
Add support for OCI charts

### LINKED ISSUE #75: Support Helm3 OCI Charts
I've been using Helm3 OCI based charts via GitLab container registry due to ease of use, but OCI based charts are not currently supported by `provider-helm`.

It would be great if OCI based charts could be supported as this _slightly_ lowers the barrier to entry due to not having to run a Helm repo (peak laziness on my behalf!)

Thoughts:

 - How to discern between a normal and an OCI based chart install (Maybe a new `Type` field which defaults to existing behaviour?)
 - How to auth with OCI registry (looks like secret already has a username and password so guessing just calling a different auth method)
 
I'm happy to look at implementing this if the time is right and we can solidify any additions to the API format.

goober:
This pull request aims to add support for pulling helm charts from OCI registries. 

In addition, the pull request also adds support for pulling charts from insecure registries by adding the option `insecureSkipTLSVerify` to the Release manifest

Since helm 3.8.0 needed some newer dependencies I decided to include a dependency upgrade in this PR as well. If it is preferable to have them as separate pull requests I can split them.

### Open questions:
* ~How should we handle insecure registries?~
* ~Should we also add support for defining the url option with `oci://` prefixed protocol? And if that is the case, should we require it to define the full tar file path?  Or support the url to be defined as `oci://localhost:5000/myrepo/mychart:2.7.0`~ 
 
Fixes #75 

I have:

Following manual tests have been performed:
* `repository` option set to a `oci://` url and also tests with version option set and without version option set.
* `url` option set to a `oci://` url with and without `:<tag>` suffix
* with and without hitting the cache

### DIFF
--- a/pkg/clients/helm/client.go
+++ b/pkg/clients/helm/client.go
@@ -51,6 +51,8 @@ const (
 	errFailedToLoadChart               = "failed to load chart"
 	errUnexpectedDirContentTmpl        = "expected 1 .tgz chart file, got [%s]"
 	errFailedToParseURL                = "failed to parse URL"
+	errFailedToLogin                   = "failed to login to registry"
+	errUnexpectedOCIUrlTmpl            = "url not prefixed with oci://, got [%s]"
 )
 
 // Client is the interface to interact with Helm
@@ -71,6 +73,7 @@ type client struct {
 	upgradeClient   *action.Upgrade
 	rollbackClient  *action.Rollback
 	uninstallClient *action.Uninstall
+	loginClient     *action.RegistryLogin
 }
 
 // ArgsApplier defines helm client arguments helper
@@ -111,6 +114,7 @@ func NewClient(log logging.Logger, restConfig *rest.Config, argAppliers ...ArgsA
 
 	pc.DestDir = chartCache
 	pc.Settings = &cli.EnvSettings{}
+	pc.InsecureSkipTLSverify = args.InsecureSkipTLSVerify
 
 	gc := action.NewGet(actionConfig)
 
@@ -119,17 +123,21 @@ func NewClient(log logging.Logger, restConfig *rest.Config, argAppliers ...ArgsA
 	ic.Wait = args.Wait
 	ic.Timeout = args.Timeout
 	ic.SkipCRDs = args.SkipCRDs
+	ic.InsecureSkipTLSverify = args.InsecureSkipTLSVerify
 
 	uc := action.NewUpgrade(actionConfig)
 	uc.Wait = args.Wait
 	uc.Timeout = args.Timeout
 	uc.SkipCRDs = args.SkipCRDs
+	uc.InsecureSkipTLSverify = args.InsecureSkipTLSVerify
 
 	uic := action.NewUninstall(actionConfig)
 
 	rb := action.NewRollback(actionConfig)
 	rb.Wait = args.Wait
 	rb.Timeout = args.Timeout
+
+	lc := action.NewRegistryLogin(actionConfig)
 
 	return &client{
 		log:             log,
@@ -139,6 +147,7 @@ func NewClient(log logging.Logger, restConfig *rest.Config, argAppliers ...ArgsA
 		upgradeClient:   uc,
 		rollbackClient:  rb,
 		uninstallClient: uic,
+		loginClient:     lc,
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/clients/helm/client.go
```diff
@@ -192,6 +203,20 @@ func (hc *client) pullChart(spec *v1beta1.ChartSpec, creds *RepoCreds, chartDir
...
 
 	pc.DestDir = chartDir
+	
+	if registry.IsOCI(spec.Repository) {
+		parsedURL, err := url.Parse(spec.Repository)
+		if err != nil {
+			return errors.Wrap(err, errFailedToParseURL)
+		}
+		var out strings.Builder
+		// TODO add support for insecure
```
goober, line 213: Where is a good place to add support for configuring insecure registries? I suppose we can add it as a property in the Release manifest but I would appreciate some suggestions for naming that property.

turkenh:
Thanks a lot for this @goober, looking pretty good. 

I would like to do some testing while reviewing this. 
Could you add/extend examples under `examples/sample` directory

pkg/clients/helm/client.go
turkenh: Hmm, good question. 
I would say somewhere under `Release` make sense like `spec.forProvider.chart.insecure`. But this would indicate that `Chart is insecure` but indeed repo/registry is insecure. 

So maybe something like

```
spec:
# rollbackLimit: 3
  forProvider:
    chart:
      name: wordpress
      repository: https://charts.bitnami.com/bitnami
      version: 9.3.19
      # Deprecated use pull.SecretRef instead
      # pullSecretRef:
      pull:
        insecure: true
        SecretRef:
          name: museum-creds
          namespace: default
```

But then we may also want to have repository and URL under pull as well 🤔

goober: We could also add it to the [Args struct](https://github.com/crossplane-contrib/provider-helm/blob/master/pkg/clients/helm/args.go ):

e.g the helm flag `--insecure-skip-tls-verify` as `InsecureSkipTLSVerify` to also add support for pulling charts from insecure repositories that are non oci registries. By that we can use the same flag to pass to the registry login helm command to indicate that the login is insecure.

```yaml
spec:
# rollbackLimit: 3
  forProvider:
    chart:
      name: wordpress
      repository: https://charts.bitnami.com/bitnami
      version: 9.3.19
    insecureSkipTLSVerify: true
    pullSecretRef:
      name: museum-creds
      namespace: default
```

goober:
@turkenh Regarding the examples, do you want public hosted charts from an OCI registry that can be used or just an example url?
I have not found any good publicly available charts that can be used in the examples.

pkg/clients/helm/client.go
turkenh: Yeah, that sounds like a better idea considering all those args are optional `flags` for `helm install` command.

goober:
> Ideally yes. Isn't it possible to use dockerhub for this purpose? For example, can we push the same workpress helm chart to dockerhub (e.g. under your own account) and add it to the example?

@turkenh Unfortunately it is not supported. See https://github.com/docker/roadmap/issues/135
However, I have added an example for pulling from a local registry that has been setup according to the following instructions form the [Helm documentation](https://helm.sh/docs/topics/registries/#running-a-registry) 

I have also updated the PR with support for insecure repositories and registries by adding the option `insecureSkipTLSVerify` to the Release manifest.

goober:
Also upgraded kubebuilder after confirmation in the following [slack thread](https://crossplane.slack.com/archives/CEF5N8X08/p1645605213253989)

turkenh (approved):
Looking great, thanks @goober!

Added two nitpicks.

pkg/clients/helm/client.go
```diff
@@ -183,16 +199,31 @@ func (hc *client) pullChart(spec *v1beta1.ChartSpec, creds *RepoCreds, chartDir
 
 	chartRef := spec.URL
 	if spec.URL == "" {
-		chartRef = spec.Name
-
-		pc.RepoURL = spec.Repository
+		if registry.IsOCI(spec.Repository) {
+			chartRef = strings.Join([]string{strings.TrimSuffix(spec.Repository, "/"), spec.Name}, "/")
```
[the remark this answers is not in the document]
goober: Since `spec.Repository` is a url with a protocol using `filepath.Join` will produce
`https:/foo.bar/bar` with a single slash for the protocol. See https://play.golang.com/p/UeoXvtYMAG1
However, I can refactor this to a separate function to make it more clear

turkenh:
@goober some check are failing, any ideas?

goober:
> @goober some checks are failing, any ideas?

@turkenh Unfortunately not. I am running `make lint` locally and it goes through. Same with `make reviewable`

goober:
@turkenh I ran `golangci-lint run` locally and found some linting errors. Upgraded golangci-lint to latest version and fixed the issues. I hope it will pass the build now

goober:
@turkenh The pull request has now been rebased after #122 and is ready for review and testing.

turkenh (approved):

hikhvar:
Hey, I like this feature! Is there an example how to pull charts from a private OCI registry? In detail: How to configure the pull secrets for that?

PR #123, opened 2022-03-08
Set config home in Dockerfile

goober:
Signed-off-by: Mathias Åhsberg <mathias.ahsberg@resurs.se>

When deploying a helm chart from an OCI registry with credentials the Release will fail with the following error:

> Warning  CannotCreateExternalResource  4s (x16 over 54s)  managed/release.helm.crossplane.io  failed to install release: failed to login to registry: mkdir /.config: permission denied

It is raised when Helm is trying to create the config folder to be used for storing auth details for the registry. 

By setting the `XDG_CONFIG_HOME ` to the `/tmp` folder in the Dockerfile we can be sure that the user running the container has permission to create the required folders.

See https://github.com/helm/helm/issues/8038 for a related issue

I have:

The PR has now been verified with installing a helm chart from a secured ECR registry in local cluster. Without the patch it will fail with the described behaviour. With the patch the Helm chart can be installed successfully.

### DIFF
--- a/cluster/images/provider-helm-controller/Dockerfile
+++ b/cluster/images/provider-helm-controller/Dockerfile
@@ -4,6 +4,7 @@ RUN apk --no-cache add ca-certificates bash
 ADD provider /usr/local/bin/crossplane-helm-provider
 
 ENV XDG_CACHE_HOME /tmp
+ENV XDG_CONFIG_HOME /tmp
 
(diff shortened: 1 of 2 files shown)


goober:
@turkenh The PR is ready for review and is a followup fix for #121

turkenh (approved):
Thanks!

PR #134, opened 2022-08-07
Support helm devel versions.

dormullor:
Helm provides the ability to specify `--devel` flag to also include development versions (alpha, beta, and release candidate releases).
Under the hood, `--devel` is equivalent to version `>0.0.0-0`.
I have added the option to specify this version in the chart and to pull the latest development version from a helm registry.

I have tested it with local Kind cluster & a private helm registry which contains an `-rc` chart.

Fixes #

I have:

- [ V] Read and followed Crossplane's [contribution process].
- [V ] Run `make reviewable` to ensure this PR is ready for review.

### DIFF
--- a/pkg/clients/helm/client.go
+++ b/pkg/clients/helm/client.go
@@ -53,7 +53,7 @@ const (
 	errFailedToParseURL                = "failed to parse URL"
 	errFailedToLogin                   = "failed to login to registry"
 	errUnexpectedOCIUrlTmpl            = "url not prefixed with oci://, got [%s]"
-	devel 							   = ">0.0.0-0"
+	devel 				   = ">0.0.0-0"
 )
 
 // Client is the interface to interact with Helm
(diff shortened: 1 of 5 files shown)


dormullor:
@turkenh Can you please review ?

pkg/clients/helm/client.go
```diff
@@ -254,7 +255,7 @@ func (hc *client) PullAndLoadChart(spec *v1beta1.ChartSpec, creds *RepoCreds) (*
 	var chartFilePath string
 	var err error
 	switch {
-	case spec.URL == "" && spec.Version == "":
+	case spec.URL == "" && (spec.Version == "" || spec.Version == devel):
```
turkenh, line 258: Hmm, what if user sets the version as `>0.0.0` and expect to have the latest _stable_ release run? 
I am wondering if the previous PR explicitly setting devel flag would be a better approach 🤔

dormullor: The previous PR doesn't work.
Apparently, the Helm client doesn't recognize the devel flag. 
That's the reason why we need to explicit provide this version.

turkenh: Thanks, sounds reasonable. 

Could you check CI, it is failing.

dorroddorrod: Fixed, can you please approve again ?

dorroddorrod (approved):

turkenh (approved):
Thanks @dormullor for bearing with the issue related to integration tests 🙌

PR #142, opened 2022-09-26
makefile: integrate xpkg publishing process

muvaf:

This adds the consumption of `xpkg` build lib so that we push XPKG to Upbound marketplace. It needs `XPKG_ACCESS_ID` and `XPKG_TOKEN` secret vars that can push to `xpkg.upbound.io/crossplane` but I don't have access to that org @hasheddan 

I borrowed most of the logic from https://github.com/crossplane-contrib/provider-aws/blob/master/Makefile

I have:

`make -j2 build.all` and see a multi-arch xpkg is built. `e2e-test` target of the PR.


hasheddan (commented):
@muvaf thanks for updating this! We are going to need some more changes for it to work fully. https://github.com/crossplane-contrib/provider-aws/pull/1460 may be useful to model after

muvaf:
The integration test fails with the following event on `Release`:
```
  Warning  CannotCreateExternalResource  2s (x8 over 27s)  managed/release.helm.crossplane.io  (combined from similar events): failed to install release: failed to pull chart: looks like "https://charts.bitnami.com/bitnami" is not a valid chart repository or cannot be reached: open /.cache/helm/repository/Bo0QkEonMrLNTmMqzq8ah5GbsCc=-index.yaml: no such file or directory
```

@hasheddan @turkenh does this ring a bell? It works as expected when I run the controller locally with `make run`.

muvaf:
Alright the problem turned out to be that distroless image didn't allow Helm to call `mkdir -p` on a root level folder, i.e. `/.cache` since `$HOME` is empty. But the error is very misleading since Helm code didn't have error handling https://github.com/helm/helm/pull/11388

Another thing is that bitnami wordpress image does not yet support ARM apparently https://github.com/bitnami/charts/issues/7305

Makefile
```diff
@@ -2,7 +2,6 @@
 PROJECT_NAME := provider-helm
 PROJECT_REPO := github.com/crossplane-contrib/$(PROJECT_NAME)
 
-PLATFORMS ?= linux_amd64 linux_arm64
```
hasheddan, line 5: @muvaf does this need to be dropped?

muvaf: I didn't see it in provider-aws and thought we probably don't need it here either. Is there a reason to keep it? The default in submodule seems to include darwin and windows as well.

hasheddan: I believe that it is present in all providers (example in [provider-aws](https://github.com/crossplane-contrib/provider-aws/blob/88bf9b6cd0545c392d3aaa247c074b9b2891d3f6/Makefile#L7)). The only issue it would cause is in image construction, but it looks like it is being skipped correctly. I think building the extra binaries here is not really needed, but I'm not strongly against it.

hasheddan (approved):
@muvaf I'll let you decide about the arch question 👍🏻 Thanks for your work here!

Makefile
muvaf: Oh I missed that. I'd prefer not to deviate, reverted that part. Thanks!

PR #153, opened 2022-11-18
Set timeout for uninstall client so delete hooks don't hang.

### LINKED ISSUE #151: Provider stops reconciling some Releases with "failed to list" and "failed to watch" *unstructured.Unstructured errors
This is  likely a client-go or helm client issue but I'm going to document it here to start since it appears in the provider-helm execution - I'll open issues in the other repositories if needed.

### What happened?
The provider will stop reconciling some Releases every few days and reports errors like the following:

```
W1112 14:33:46.500749       1 reflector.go:424] k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: failed to list *unstructured.Unstructured: Unauthorized
E1112 14:33:46.500781       1 reflector.go:140] k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: Failed to watch *unstructured.Unstructured: failed to list *unstructured.Unstructured: Unauthorized
W1112 14:34:44.496384       1 reflector.go:424] k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: failed to list *unstructured.Unstructured: Unauthorized
E1112 14:34:44.496418       1 reflector.go:140] k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: Failed to watch *unstructured.Unstructured: failed to list *unstructured.Unstructured: Unauthorized
W1112 14:35:22.419968       1 reflector.go:424] k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: failed to list *unstructured.Unstructured: Unauthorized
E1112 14:35:22.420001       1 reflector.go:140] k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: Failed to watch *unstructured.Unstructured: failed to list *unstructured.Unstructured: Unauthorized
```
This appears to happen to Releases that are marked for deletion.  Removing the helm release from the target cluster manually does not fix the problem - the provider pod has to be restarted to resume reconciliation.

I suspect that the problem is related to delete hook processing in the helm client, as that's the only path I can see that invokes this code - the provider calls uninstall in the helm client, which executes the delete hooks and waits/watches for them to complete.

The errors are being reported from here:

https://github.com/kubernetes/client-go/blob/master/tools/cache/reflector.go#L425

and here:

https://github.com/kubernetes/client-go/blob/master/tools/cache/reflector.go#L140

which report the errors but do do anything to stop the controller that is doing the watch.

That would seem to require overriding the DefaultWatchErrorHandler as in the example here:

https://github.com/kubernetes/kubernetes/pull/87329/files#diff-f4e24e0a1a1c8de4b31a42ebfa2218f19a5d70fb99779b4af6e04a024fabe3d5R343

which would cause the informer controller to exit.

That appears to require changes in client-go tools/cache/controller.go to allow NewIndexerInformerWatcher to set the the watch controller's watchErrorHandler before the controller starts.

I started a Slack conversation for this in #sig-api-machinery so we'll see what happens.

### How can we reproduce it?
I have not found the exact scenario to reproduce the problem, though it appears to be related to delete handling.

### What environment did it happen in?
Crossplane version: 1.10.1

* Cloud provider or hardware configuration - AWS
* Kubernetes version (use `kubectl version`) - 1.21
* Kubernetes distribution (e.g. Tectonic, GKE, OpenShift) - EKS

bobh66:
Signed-off-by: Bob Haddleton <bob.haddleton@nokia.com>

Set the uninstall client timeout and wait parameters to the specified args,

The default timeout is null which causes the helm client to wait forever for delete hooks which may never complete if the cluster has been deleted.

Fixes #151 

I have:

Soaked in our development environment for a week, which usually sees the problem multiple times. 

### DIFF
--- a/pkg/clients/helm/client.go
+++ b/pkg/clients/helm/client.go
@@ -133,6 +133,8 @@ func NewClient(log logging.Logger, restConfig *rest.Config, argAppliers ...ArgsA
 	uc.InsecureSkipTLSverify = args.InsecureSkipTLSVerify
 
 	uic := action.NewUninstall(actionConfig)
+	uic.Wait = args.Wait
+	uic.Timeout = args.Timeout
 
 	rb := action.NewRollback(actionConfig)
 	rb.Wait = args.Wait


turkenh (approved):
Thanks a lot @bobh66 👍

PR #160, opened 2022-11-29
Optionally skip 'partOfRelease' check

### LINKED ISSUE #99: Getting connection detail from manually generated secret

I can reference a secret that is generated by provider-helm, like the following example sited from [example](https://github.com/crossplane-contrib/provider-helm/blob/master/examples/sample/release.yaml). But if that secret is generated manually, it can't be referenced, the generated credential would be empty. 
``` yaml
    - apiVersion: v1
      kind: Secret
      name: wordpress-example
      namespace: crossplane-svc
      fieldPath: data.wordpress-password
      toConnectionSecretKey: password
```

Is there any chance we can reference secrets that are not generated by provider-helm? (or any workarounds?) There's scenarios where we have to generate certs manually, and deliver it to end application via crossplane. Thanks!

Ranger-X:
Add ability to skip 'is part of release' check for manually (non-helm managed) created resource when generating ConnectionDetails.

PS. Field **skipPartOfReleaseCheck** added to _ConnectionDetail_ struct instead of _ReleaseParameters_ (which makes more sense IMHO) because there is we cannot (or can but I'm do not know how? :-) ) access _ReleaseParameters_ from _connectionDetails_ function.

Fixes #99 

I have:

Added 'Success_NotPartOfReleaseAndSkipPartOfReleaseCheck' test to observe_test.go.

### DIFF
--- a/apis/release/v1beta1/types.go
+++ b/apis/release/v1beta1/types.go
@@ -124,6 +124,8 @@ type ReleaseStatus struct {
 type ConnectionDetail struct {
 	v1.ObjectReference    `json:",inline"`
 	ToConnectionSecretKey string `json:"toConnectionSecretKey,omitempty"`
+	// SkipPartOfReleaseCheck skips check for meta.helm.sh/release-name annotation.
+	SkipPartOfReleaseCheck bool `json:"skipPartOfReleaseCheck,omitempty"`
 }
 
 // +kubebuilder:object:root=true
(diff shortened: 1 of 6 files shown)


turkenh (approved):
Thank you @Ranger-X!
Could you also extend the example [here](https://github.com/crossplane-contrib/provider-helm/blob/2153863b86bd8391bb14f2240d0232c357bca54f/examples/sample/release.yaml#L45) to make it move discoverable?

Ranger-X:
> Thank you @Ranger-X! Could you also extend the example [here](https://github.com/crossplane-contrib/provider-helm/blob/2153863b86bd8391bb14f2240d0232c357bca54f/examples/sample/release.yaml#L45) to make it move discoverable?

Yep. Done :-)

PR #196, opened 2023-08-10
Add Proxy Support

### LINKED ISSUE #195: Add Proxy Support

Due to transetive peering issues within the GCP we need to use a proxy to connect to the control plane. This feature is currently not available within in the helm provider.

Add the possibility to connect via a proxy.

oskisk:

Adds a proxy support for connecting to the Kubernetes API server.

Fixes #195

I have:

Tested it manually by installing a helm chart via a proxy.

### DIFF
--- a/pkg/controller/release/release.go
+++ b/pkg/controller/release/release.go
@@ -18,6 +18,8 @@ package release
 
 import (
 	"context"
+	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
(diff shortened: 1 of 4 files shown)


pkg/controller/release/release.go
```diff
@@ -202,6 +205,14 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
 		}
 	}
 
+	if proxy := p.Spec.Proxy; proxy != "" {
```
turkenh, line 208: Wouldn't it work by passing proxy/no_proxy environment variables via a ControllerConfig ? https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml

oskisk: This would work when we only need to connect to a single proxy instance per controller.

In our case we use one crossplane/helm provider instance to manage multiple clusters in different vpc networks. Therefore we need to connect to a different proxy instance depending of the cluster/vpc.

oskisk: @turkenh

Pocokwins: Hi @turkenh ,

Is there any update on this PR?

PR #207, opened 2023-12-20
[WIP] feat: import kubelogin library mode

bcho:

Fixes #

I have:


go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/crossplane-contrib/provider-helm
 
-go 1.20
```
bcho, line 3: is it okay to bump to 1.21+ ?

go.mod
```diff
@@ -1,9 +1,9 @@
...
+go 1.21.3
 
 require (
-	github.com/Azure/kubelogin v0.0.34
```
bcho, line 6: NOTE: once we confirm the change, we will cut a release in kubelogin side.

pkg/clients/azure/azure.go
```diff
@@ -18,26 +18,38 @@ const (
...
+	return opts, nil
+}
+
 func WrapRESTConfig(_ context.Context, rc *rest.Config, credentials []byte, _ ...string) error {
 	m := map[string]string{}
 	if err := json.Unmarshal(credentials, &m); err != nil {
 		return err
 	}
 
-	fs := pflag.NewFlagSet("kubelogin", pflag.ContinueOnError)
```
bcho, line 29: @erhancagirici I am less sure how to crossplane consume this rest config: are we expecting the input config is returned from an AAD enabled AKS cluster? Because using kubelogin's SP login method, it's required to have `server-id` set.

bcho (commented):
@erhancagirici kubelogin has merged the library mode change, this is the draft pr for showing how to use the library mode. A few things needs to confirm with your side's usage. Please see the comments.

go.mod
bcho: reverted

pkg/clients/azure/azure.go
```diff
@@ -18,26 +18,38 @@ const (
...
+	kubeloginCLIFlagServerID = "server-id"
 )
 
+func kubeloginTokenOptionsFromRESTConfig(rc *rest.Config) (*token.Options, error) {
+	opts := &token.Options{}
+
+	// opts are filled according to the provided args in the execProvider section of the kubeconfig
+	// we are parsing serverID from here
+	// add other flags if new login methods are introduced
+	fs := pflag.NewFlagSet("kubelogin", pflag.ContinueOnError)
```
erhancagirici, line 31: Looks like this complains while parsing parse kubelogin flags, due to undefined flags being present in the supplied arguments. I was able to overcome this with the following configuration on `pflag`
```suggestion
	fs := pflag.NewFlagSet("kubelogin", pflag.ContinueOnError)
	fs.ParseErrorsWhitelist = pflag.ParseErrorsWhitelist{UnknownFlags: true}
```

Of course, as I described in my [previous comment](https://github.com/crossplane-contrib/provider-helm/pull/207#discussion_r1447110675), ideally I believe it would be more robust if we could parse this through some `kubelogin`-maintained utility func and directly obtain all library related flags. Otherwise, we need to maintain the same list of `kubelogin` flags here.

erhancagirici (changes requested):
@bcho thanks a ton for this pull request and upstream `kubelogin` changes, much appreciated!
Left a comment regarding the flag parsing and code change suggestion. With the changed suggestion I was able to test and validate that it is working.

pkg/clients/azure/azure.go
bcho: thanks for your review, I agree with you on the flags parsing pain here. I created an issue to track it: https://github.com/Azure/kubelogin/issues/394 and will see how to expose it in library mode. We can follow up after the change in kubelogin side.

PR #210, opened 2024-02-02
feat(in-cluster): update example to use native crossplane way

haarchri:
use crossplane native way to grant a static service account the needed cluster-admin rights

Fixes #

I have:


turkenh (approved):
Thanks!

PR #223, opened 2024-05-24
Consume kube client package from provider-kubernetes & new identity modes for auth

turkenh:

This PR consumes the exported kube client package in https://github.com/crossplane-contrib/provider-kubernetes/pull/251, by dropping the duplicate implementation here. This also introduces support for authenticating with the following identities:

- AzureWorkloadIdentityCredentials
- UpboundToken

I have:

See the steps in, adapted for provider-helm https://github.com/crossplane-contrib/provider-kubernetes/pull/251


go.mod
```diff
@@ -1,75 +1,74 @@
 module github.com/crossplane-contrib/provider-helm
 
-go 1.22.0
+go 1.22.1
 
-toolchain go1.22.1
+replace github.com/crossplane-contrib/provider-kubernetes => github.com/turkenh/provider-kubernetes v0.0.0-20240523170107-bea981f8819d
```
turkenh, line 5: TODO: remove this when https://github.com/crossplane-contrib/provider-kubernetes/pull/251 merged.

lsviben:
Do we want to move that kube client package somewhere generic? Not saying now, but maybe some future step? Although not sure if there are other providers that would use it except helm/kubernetes, so maybe not worth it.

turkenh:
> Do we want to move that kube client package somewhere generic? Not saying now, but maybe some future step? Although not sure if there are other providers that would use it except helm/kubernetes, so maybe not worth it.

We have been duplicating the code here and there since we couldn't find a good home for this code till today. We could move it to crossplane-runtime but that would mean introducing dependencies like azure kubelogin or upbound-up-sdk etc. 
What I am doing here is better than duplicating the code, which became hard to maintain as the number of auth methods increased while not being perfect. I am not aware of any other providers needing this, so probably already good enough for now.

PR #262, opened 2025-01-03
fix: remove omitempty in invalid json tag combination

kkendzia:

As mentioned in https://github.com/crossplane/function-sdk-go/issues/161 and further https://github.com/go-json-experiment/json/pull/39 "inline,omitempty" is an invalid combination of json tags. This leads to issues in marshaling DataKeySelector. 

I have:

We have tested the modified types.go locally in our e2e tests. This shouldn't be a breaking change as it wasn't working in crossplane functions until now. Also CRDs have not changed.

Would appreciate a 0.20 or even 0.19 backport


lsviben (approved):
Makes sense, thanks for the fix!

PR #298, opened 2025-08-22
Add a field to specify namespaces for `releases.helm.m.crossplane.io`

### LINKED ISSUE #297: Allow the namespaced Release managed resource to install the Helm Release in a different namespace
I'm trying to use the namespaced release but the provider is creating the release in a different namespace from the deployment.

I have a `release.helm.m.crossplane.io` object for external secrets manager in a "hub" cluster in a namespace called `ty`. The `release.helm.m.crossplane.io` is using a provider config with a kubeconfig secret for a remote cluster called "test-cluster". I would like the release secret and external secrets manager to be created in the `external-secrets` namespace on the remote cluster. However, everything is created in the `ty` namespace on the remote cluster which doesn't exist.

I would like an option similar to the `spec.forProvider.namespace` field in the cluster scoped release add to the namespace scoped release ( or an equivalent to `helm install -n external-secrets`)?

tydanny:

Fixes #297 

This PR adds a namespace field to the `forProvider` section of the spec of the namespaced Release object. This PR also changes the behavior of the namespaced release controller to use the Managed Release's namespace as the helm client's namespace by default or use the one specified in the new field if set.

I was unsure what the default behavior should be so I opted to maintain the current behavior by default.

I have:

I tested these changes by running them locally against a "hub" cluster with a provider config for a "remote" cluster. I then created the example WordPress release with `spec.forProvider.namespace` set to `wordpress`. The release was successfully created in the `wordpress` namespace in the remote cluster.

I also tested the default behavior by creating an example release in the `wordpress2` namespace of the hub cluster using the same config provider and without specifying `spec.forProvider.namespace`. I then verified that the release was created in the remote cluster in the `wordpress2` namespace.


ig-bwink:
I tested this PR as I had the same issue that namespaced MRs couldn't specify a namespace for remote installation. This PR fixed that but I also noticed that it does not create namespaces, which appears to be the default behavior of the cluster-scoped MRs. Would it be possible to add that back as well if a namespace is defined?

bobh66:
It looks like the namespaced `Release` implementation assumes that the chart will be deployed in the local namespace, which means it doesn't need to be created.  @sergenyalcin is the namespaced `Release` supposed to be able to install into a remote cluster, with a different namespace that may need to be created?

tydanny:
> I tested this PR as I had the same issue that namespaced MRs couldn't specify a namespace for remote installation. This PR fixed that but I also noticed that it does not create namespaces, which appears to be the default behavior of the cluster-scoped MRs. Would it be possible to add that back as well if a namespace is defined?

I'm on vacation this week but I will add this to my PR once I'm back home.

jeanduplessis:
@Scrumplex Release is out now: https://github.com/crossplane-contrib/provider-helm/releases/tag/v1.0.2

PR #323, opened 2026-02-20
Fix ignored `secretRef.namespace` in namespaced ProviderConfigs 

aerfio:

Previous implementation overrides the `secretRef.namespace` field of the ProviderConfig, it uses the namespace of the namespaced Release object instead. I treat it as a bug, because this field is effectively ignored and useless in this case. We'd rather not copy the secret with kubeconfig to each namespace where it's consumed.

I have:

Added new unit test for the `resolveProviderConfigModern` function.

### DIFF
--- a/pkg/clients/helm/helm.go
+++ b/pkg/clients/helm/helm.go
@@ -88,7 +88,6 @@ func resolveProviderConfigModern(ctx context.Context, crClient kclient.Client, m
 	var pcSpec kconfig.ProviderConfigSpec
 	switch pc := pcObj.(type) {
 	case *namespacedv1beta1.ProviderConfig:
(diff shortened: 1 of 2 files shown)


jonasz-lasut:
This behavior is consistent with all other official providers (aws, azure(ad), gcp) and it was done on purpose to limit the capabilities of namespace scoped provider configs, most kubernetes controllers limit secret references to the same namespace.

I think more correct solution would be to remove `spec.credentials.secretRef.namespace` from namespaced `ProviderConfig` because the current UX is confusing

aerfio:
@jonasz-lasut I adjusted this provider to behave the same way as provider-kubernetes, in which this feature works. I'm gonna close this PR then

PR #324, opened 2026-02-24
feat: support installing charts with digest from OCI registries

stevendborrelli:

- Adds support for SHA digests, as documented in <https://helm.sh/docs/v3/topics/registries#installing-charts-with-digest>
- Updates golangci-lint to 2.10.1
- Add e2e tests for digest testing using podinfo chart

Fixes #

I have:

e2e tests now include digest testing for cluster and namespaced scoped resources:

```shell
UPTEST_EXAMPLE_LIST ?= "examples/cluster/sample/release.yaml,examples/namespaced/sample/release.yaml,examples/cluster/sample/release-oci-digest.yaml,examples/namespaced/sample/release-oci-digest-url.yaml"
```

Installed example manifests into a cluster using a test build:

```shell
$ kubectl get -f examples/namespaced/sample/release-oci-digest.yaml         
NAME                     CHART     VERSION   SYNCED   READY   STATE      REVISION   DESCRIPTION        AGE
podinfo-digest-example   podinfo   6.10.2    True     True    deployed   1          Install complete   45m
```

```shell
$ kubectl get -f examples/namespaced/sample/release-oci-digest-url.yaml
NAME                         CHART     VERSION   SYNCED   READY   STATE      REVISION   DESCRIPTION        AGE
podinfo-digest-url-example   podinfo   6.10.2    True     True    deployed   1          Install complete   7s
```


pkg/clients/helm/client.go
```diff
@@ -269,54 +285,196 @@ func (hc *client) PullAndLoadChart(mg resource.Managed, creds *RepoCreds) (*char
...
+					hc.log.Debug("Cache hit: digest matches", "path", cachedPath, "digest", d)
+					chartFilePath = cachedPath
+				} else {
+					hc.log.Debug("Cache digest mismatch, will pull fresh", "path", cachedPath, "requestedDigest", d, "cachedDigest", cachedDigest)
+					chartFilePath = ""
+				}
+			} else {
+				// No version to construct cache path, must pull
+				chartFilePath = ""
+			}
```
phisco: Not: maybe make this a helper function so that you can return instead of all the nesting?

pkg/clients/helm/client.go
```diff
@@ -269,54 +285,196 @@ func (hc *client) PullAndLoadChart(mg resource.Managed, creds *RepoCreds) (*char
...
+					hc.log.Debug("Cache digest mismatch, will pull fresh", "path", cachedPath, "requestedDigest", chartDigest, "cachedDigest", cachedDigest)
+					chartFilePath = ""
+				}
+			} else {
+				// No name/version to construct cache path, must pull
+				chartFilePath = ""
+			}
+		} else {
+			chartFilePath = resolveChartFilePath(chartName, chartVersion)
+		}
```
phisco, line 415: And reuse it here too?

pkg/clients/helm/client.go
stevendborrelli: I've factored out the code into a helper `resolveCachedChartPathWithDigest`.

pkg/clients/helm/client.go
stevendborrelli: converted to `ensureChartCached`()`

pkg/controller/cluster/release/release.go
```diff
@@ -283,16 +283,24 @@ func (e *helmExternal) deploy(ctx context.Context, cr *v1beta1.Release, action d
...
 		cr.Spec.ForProvider.Chart.Version = chart.Metadata.Version
+		needsUpdate = true
+	}
+	// Note: Digest is NOT late-initialized because:
+	// 1. Chart metadata doesn't contain OCI digest
+	// 2. Digest is user-specified for immutability
+	// 3. Helm verifies digest during pull
+
+	if needsUpdate {
 		if err := e.localKube.Update(ctx, cr); err != nil {
```
erhancagirici, line 305: nit: not introduced in this PR, but this update seems to be not checking for `LateInitialize` management policy.

pkg/controller/cluster/release/release.go
erhancagirici, line 240: (possible nit) question:
This is not in the changed files of PR atm, so commenting here:
Should  `isUpToDate()` implementation updated to consider newly introduced `digest` field?

AFAIU, if `spec.forProvider.chart.digest` is changed in the manifest (and no `name` or `version` changes), we would still return `upToDate` true.
https://github.com/crossplane-contrib/provider-helm/blob/2059527d29977ede8ca34f47e6eee8412c1d8f30/pkg/controller/cluster/release/observe.go#L85-L98

apis/cluster/release/v1beta1/types.go
erhancagirici: nit: IIUC, this might not be complete after introduction of `digest`, possibly something like
```suggestion
	// Version of Helm chart. If not set, gets late initialized: 
	//   - with corresponding version when `digest` is specified 
	//   - with latest version otherwise
```

examples/cluster/sample/release-oci-digest.yaml
```diff
@@ -0,0 +1,19 @@
...
+    chart:
+      # Pull repository + name + digest
+      name: podinfo
+      # Digest provides immutable chart pinning for enhanced supply chain security
+      # Only supported for OCI registries
+      # When digest is specified, the provider validates cached charts against the digest
+      # and only re-downloads if the digest doesn't match, reducing network traffic
+      digest: sha256:c56f4d760bc9da702f231f37fcec89c66b0993f0cb91446f86d014b133c6693f
+      repository: oci://ghcr.io/stefanprodan/charts
+      version: 6.10.2
```
erhancagirici, line 16: very nit: maybe an extra example manifest without the version (digest only), to showcase that it is a valid configuration.

pkg/controller/cluster/release/release.go
```diff
@@ -283,16 +283,24 @@ func (e *helmExternal) deploy(ctx context.Context, cr *v1beta1.Release, action d
...
 	if cr.Spec.ForProvider.Chart.Name == "" {
 		cr.Spec.ForProvider.Chart.Name = chart.Metadata.Name
-		if err := e.localKube.Update(ctx, cr); err != nil {
-			return errors.Wrap(err, errFailedToSetName)
-		}
+		needsUpdate = true
 	}
 	if cr.Spec.ForProvider.Chart.Version == "" {
 		cr.Spec.ForProvider.Chart.Version = chart.Metadata.Version
+		needsUpdate = true
```
erhancagirici, line 299: Q: could you clarify what is the expected UX around the following scenario?

1. User creates a release with `spec.forProvider.chart.digest` only (no `version` set).
2. `version` gets late initialized
3. The consumer wants to upgrade their chart, changes `spec.forProvider.chart.digest` with new value, without `version`
Now: 
do we want version to be "late-initialized" again?
  - this codepath won't be handling that IIUC
alternatively, do we expect users to keep `version` and `digest` in sync?
  - IIUC, the next reconciliation would deliberately fail the validation for the digest mismatch, so the consumer will be notified

A follow up question, should the users have the ability to opt out of the late-init here and purely manage `digest`s in their desired manifests?

erhancagirici, line 299: Q: could you clarify what would be the intended/expected UX around the following scenario?

1. The consumer creates a release with `digest` only (no version set).
2. `version` gets late initialized
3. Consumer wants to upgrade their chart, changes `spec.forProvider.chart.digest` without `version`
  - e.g. Consumer wants to only manage digests
Now: 
- "late-initialized" again?
  - this codepath won't be handling that 
- or, do we expect this to ignore version here?
  - maybe opt out of late init here?
- expect consumers to update `version` when they update`digest`?
  - might be problematic with gitops flows where digest is specified only

erhancagirici (commented):
@stevendborrelli thanks for handling this 🙌  Looks good to me overall, just left some comments for clarification and consideration

pkg/controller/cluster/release/release.go
stevendborrelli: I'm going to remove the late-initialization for version. It doesn't make sense to populate it when the digest is set. We can store the version in the status.

pkg/controller/cluster/release/release.go
stevendborrelli: I've updated the code to check for digest in `isUpToDate()`

examples/cluster/sample/release-oci-digest.yaml
stevendborrelli: Added:

- [examples/cluster/sample/release-oci-digest-only.yaml](examples/cluster/sample/release-oci-digest-only.yaml)
- [examples/namespaced/sample/release-oci-digest-only.yaml](examples/cluster/sample/release-oci-digest-only.yaml)

pkg/controller/cluster/release/release.go
stevendborrelli: I've removed late-initializing the version when a digest is provided.

apis/cluster/release/v1beta1/types.go
stevendborrelli: I've opted to store the version in status so that we don't have to lateInitialize the version when a digest is specified.

pkg/controller/cluster/release/release.go
stevendborrelli: I've updated the code to check for lateInit.

pkg/controller/cluster/release/release.go
```diff
@@ -225,6 +226,8 @@ func (e *helmExternal) Observe(ctx context.Context, mg resource.Managed) (manage
 	}
 
 	cr.Status.AtProvider = generateObservation(rel)
+	// Store the digest in status for drift detection
+	cr.Status.AtProvider.Digest = cr.Spec.ForProvider.Chart.Digest
```
erhancagirici: If I am not missing anything, setting digest to `status.atProvider` here would cause the comparison to be always equal in the following `isUpToDate` check. This results in failing to detect the drift when digest is updated.
```suggestion
```
I suppose `status.atProvider.digest` should be only set after a successful deploy operation (which is already done in this PR 👍 )

We can add a test case for this as well.

Same applies for the namespaced controller.

pkg/controller/namespaced/release/release.go
```diff
@@ -230,6 +231,8 @@ func (e *helmExternal) Observe(ctx context.Context, mg resource.Managed) (manage
 	}
 
 	cr.Status.AtProvider = generateObservation(rel)
+	// Store the digest in status for drift detection
+	cr.Status.AtProvider.Digest = cr.Spec.ForProvider.Chart.Digest
```
erhancagirici: same as https://github.com/crossplane-contrib/provider-helm/pull/324/changes#r3358569643

pkg/clients/helm/client.go
erhancagirici: nit: not initially introduced in this PR particularly, this can be a follow-up. Just wanted to note.
Function name seems to be a bit confusing, this helper does not only pull "latest" chart.

pkg/clients/helm/client.go
```diff
@@ -269,54 +351,114 @@ func (hc *client) PullAndLoadChart(mg resource.Managed, creds *RepoCreds) (*char
...
+	case chartUrl == "" && (chartVersion == "" || chartVersion == devel) && chartDigest == "":
+		// No URL, no version, no digest -> pull latest
 		chartFilePath, err = hc.pullLatestChartVersion(chartUrl, chartName, chartVersion, chartRepo, creds)
 		if err != nil {
 			return nil, err
 		}
 	case registry.IsOCI(chartUrl):
-		u, v, err := resolveOCIChartVersion(chartUrl)
+		// OCI URL provided
+		u, v, d, err := resolveOCIChartVersionAndDigest(chartUrl)
```
erhancagirici: nit: This can be a follow-up, but wanted to note here:
By design, chart URL is intended to be the source of truth when specified, overriding the dedicated spec parameters `spec.forProvider.chart.digest` and `spec.forProvider.chart.version`. This is also noted in the docstring of `spec.forProvider.chart.URL` 👍 

Though, we don't report mismatches when there are conflicting information. UX-wise, this can be misleading, as we ignore those values.

pkg/clients/helm/client.go
```diff
@@ -41,11 +44,16 @@ import (
...
-	chartCache        = "/tmp/charts"
+	helmDriverSecret = "secret"
+	// releaseMaxHistory is the maximum number of entries Helm will keep in
+	// release history. We just set a reasonable default for our use case.
 	releaseMaxHistory = 20
 )
 
+// chartCache is the directory where pulled chart tarballs are stored. It is
+// mutable in tests so that they can override it with a temporary location.
+var chartCache = "/tmp/charts"
```
erhancagirici, line 52: nit: instead of a global mutable variable, we can consider making this a member of `helm.client` with a default, and override in tests only.

pkg/clients/helm/client.go
```diff
@@ -364,27 +506,144 @@ func (hc *client) Uninstall(release string) error {
...
+// The digest string is returned in the OCI format "sha256:<hex>".
+// The filePath is sanitized to only use the base filename, preventing path traversal.
+func computeFileDigest(filePath string) (string, error) {
+	// Determine the base directory from the input path
+	// In production: chartCache (/tmp/charts)
+	// In tests: the directory containing the file
+	baseDir := filepath.Dir(filePath)
+	if baseDir == "." || baseDir == "" {
+		baseDir = chartCache
+	}
```
erhancagirici: nit: IMHO this function should not have a base dir heuristic, and should have a cleaner interface regarding expected paths (absolute vs relative). In the current state of the PR all calls are absolute path. For the sake of simplicity in the code, I think we should follow a consistent pattern, like:
- always accept absolute paths, assert and error-out if we get an unexpected base dir
- or this function should only accept filenames, assert/sanitize that, and always open from predetermined base dir.

This is also relevant for other path-related changes in this PR.

pkg/clients/helm/client.go
```diff
@@ -259,8 +280,69 @@ func (hc *client) login(chartUrl, chartRepo string, creds *RepoCreds, insecure b
...
+		pulledName, err := getChartFileName(tmpDir)
+		if err != nil {
+			return "", err
+		}
+
+		// Atomically move pulled artifact into cache using safe path construction
+		dstPath := safePath(chartCache, chartFileName)
+		srcPath := safePath(tmpDir, pulledName)
+		// G703: safePath() sanitizes with filepath.Base() preventing traversal
+		if err := os.Rename(srcPath, dstPath); err != nil { //nolint:gosec
```
erhancagirici: nit: to be consistent with `os.OpenRoot`, this maybe should utilize `root.Rename` with root-relative paths.

pkg/controller/cluster/release/release.go
```diff
@@ -283,16 +286,31 @@ func (e *helmExternal) deploy(ctx context.Context, cr *v1beta1.Release, action d
...
+		// Note: When digest is used, the actual deployed version is stored in status.atProvider.version
+		// for observability, but the spec is not modified to avoid drift in GitOps workflows
 	}
-	if cr.Spec.ForProvider.Chart.Version == "" {
-		cr.Spec.ForProvider.Chart.Version = chart.Metadata.Version
+
+	if needsUpdate {
 		if err := e.localKube.Update(ctx, cr); err != nil {
-			return errors.Wrap(err, errFailedToSetVersion)
+			return errors.Wrap(err, "failed to update chart spec with late-initialized values")
```
erhancagirici: nit: might be better to externalize this error string

pkg/controller/cluster/release/release.go
erhancagirici: nit: these two error messages become unused with the change at
https://github.com/crossplane-contrib/provider-helm/pull/324/changes/BASE..e787fedb502be92bdbdbb54db04afdba0a3943fd#r3373416561
and collapsed into one.

```suggestion
errFailedToLateInitialize            = "failed to update chart spec with late-initialized values"
```

examples/cluster/sample/release-oci-digest-url.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: podinfo-digest-url-example
+spec:
+  forProvider:
+    chart:
+      # Using digest directly in the URL (OCI format)
+      # This ensures the exact chart is pulled regardless of tags
+      # The provider validates cached charts against the digest and reuses them when possible
+      #repository: oci://ghcr.io/stefanprodan/charts/podinfo
+      #digest: sha256:c56f4d760bc9da702f231f37fcec89c66b0993f0cb91446f86d014b133c6693f
```
erhancagirici: nit: looks like leftover commented-out values

examples/namespaced/sample/release-oci-digest-url.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: podinfo-digest-url-example
+  namespace: crossplane-system
+spec:
+  forProvider:
+    chart:
+      # Using digest directly in the URL (OCI format)
+      # This ensures the exact chart is pulled regardless of tags
+      # The provider validates cached charts against the digest and reuses them when possible
+      #repository: oci://ghcr.io/stefanprodan/charts/podinfo
+      #digest: sha256:c56f4d760bc9da702f231f37fcec89c66b0993f0cb91446f86d014b133c6693f
```
erhancagirici: nit: looks like leftover commented-out values

examples/cluster/sample/release-oci-digest-only.yaml
```diff
@@ -0,0 +1,19 @@
...
+    chart:
+      # Pull repository + name + digest (version is optional and not late-initialized)
+      name: podinfo
+      # When digest is specified:
+      # 1. Chart is pulled using the digest for immutable, secure deployment
+      # 2. Version in spec is optional and NOT late-initialized to avoid GitOps drift
+      # 3. Actual deployed version is available in status.atProvider.version for observability
+      # This is ideal for GitOps workflows that manage only digests
+      digest: sha256:c56f4d760bc9da702f231f37fcec89c66b0993f0cb91446f86d014b133c6693f
+      repository: oci://ghcr.io/stefanprodan/charts
```
erhancagirici: note: these oci targets add internet dependency to uptest
Any chance we could depend on a oci repo that we have more control over its content? maybe some chart from xp repositories, or setup a local OCI repo for test purposes.

pkg/clients/helm/client.go
```diff
@@ -269,54 +351,114 @@ func (hc *client) PullAndLoadChart(mg resource.Managed, creds *RepoCreds) (*char
...
+		u, v, d, err := resolveOCIChartVersionAndDigest(chartUrl)
 		if err != nil {
 			return nil, err
 		}
 
-		if v == "" {
+		switch {
+		case d != "":
+			// Digest in URL: check if cached version matches digest
+			chartFilePath = resolveCachedChartPathWithDigest(path.Base(u.Path), v, d, hc.log)
```
erhancagirici: can be a follow-up: IIUC digest-only (no version) URLs will never cache the charts and repull. We should consider caching these as well.

erhancagirici (approved):

PR #328, opened 2026-04-01
Add TakeOwnership Option for Helm Release Management

stevendborrelli:

Add TakeOwnership Option for Helm Release Management

Upstream Helm PR <https://github.com/helm/helm/pull/12876>

This PR includes the SHA support code from #324. It will remain in draft until that is merged. 

## Summary

Add a `takeOwnership` boolean field to Release resources that allows provider-helm to ignore Helm annotation checks and take ownership of existing Helm releases. This enables Crossplane to manage Helm releases that were initially deployed outside of Crossplane.

## Problem

When adopting existing Helm releases into Crossplane management, Helm's default behavior prevents taking over releases that lack the proper Crossplane annotations. This creates a barrier for teams migrating existing infrastructure to Crossplane-managed workflows.

## Solution

Introduce a `takeOwnership` field that bypasses Helm's annotation checks, allowing Crossplane to adopt and manage pre-existing Helm releases.

## API Changes

Added `TakeOwnership` boolean field to both cluster-scoped and namespaced Release APIs:

```yaml
kind: Release
metadata:
  name: my-release
spec:
  forProvider:
    chart:
      name: mychart
      repository: https://charts.example.com
      version: 1.2.3
    namespace: default
    takeOwnership: true  # New field
  providerConfigRef:
    name: helm-provider
```

## Implementation Details

- Added `TakeOwnership` field to `ReleaseParameters` in cluster and namespaced APIs
- Extended `Args` struct to pass the configuration through the Helm client
- Configured install and upgrade clients to respect the `TakeOwnership` setting
- Wired field through controller to apply to Helm operations

## Test Coverage

Added unit tests:
- Test case with `takeOwnership: true`
- Test case with `takeOwnership: false`
- Validation that the field properly flows through the configuration chain

## Use Cases

1. **Migration**: Adopt existing Helm releases deployed manually or via other tools
2. **Disaster Recovery**: Restore Crossplane management after manual intervention
3. **Multi-Tool Workflows**: Manage releases that may be temporarily operated by other automation

## Notes

- This field is optional and defaults to `false` to maintain safe, non-invasive behavior
- When enabled, bypasses Helm's annotation-based ownership checks
- Works with both Install and Upgrade operations
- Generated CRDs automatically include the new field

I have:

### DIFF
--- a/apis/cluster/release/v1beta1/types.go
+++ b/apis/cluster/release/v1beta1/types.go
@@ -106,6 +106,9 @@ type ReleaseParameters struct {
 	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
 	// PlainHTTP uses insecure HTTP connections for the chart download
 	PlainHTTP bool `json:"plainHTTP,omitempty"`
+	// TakeOwnership ignore the check for helm annotations and take ownership
+	// of the existing resources.
+	TakeOwnership bool `json:"takeOwnership,omitempty"`
 }
 
 // ReleaseObservation are the observable fields of a Release.
--- a/apis/cluster/release/v1beta1/types.go
+++ b/apis/cluster/release/v1beta1/types.go
@@ -106,8 +106,11 @@ type ReleaseParameters struct {
 	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
 	// PlainHTTP uses insecure HTTP connections for the chart download
 	PlainHTTP bool `json:"plainHTTP,omitempty"`
-	// TakeOwnership ignore the check for helm annotations and take ownership
-	// of the existing resources.
+	// TakeOwnership ignores Helm ownership validation and adopts pre-existing releases.
+	// This is a ONE-TIME operation: after the first successful deployment, the flag is recorded
+	// in status.atProvider.ownershipTaken and subsequent reconciles use normal Helm validation.
+	// This prevents silent adoption of unrelated resources during chart upgrades.
+	// Use this field to migrate manually-deployed Helm releases into Crossplane management.
 	TakeOwnership bool `json:"takeOwnership,omitempty"`
 }
 
@@ -120,6 +123,10 @@ type ReleaseObservation struct {
 	Digest string `json:"digest,omitempty"`
 	// Version is the actual deployed chart version.
 	Version string `json:"version,omitempty"`
+	// OwnershipTaken indicates that spec.forProvider.takeOwnership was used for initial adoption.
+	// Once set to true, subsequent reconciles use normal Helm validation instead of takeOwnership,
+	// preventing silent adoption of unrelated resources during upgrades.
+	OwnershipTaken bool `json:"ownershipTaken,omitempty"`
 }
 
 // A ReleaseSpec defines the desired state of a Release.
--- a/pkg/controller/cluster/release/release.go
+++ b/pkg/controller/cluster/release/release.go
@@ -154,6 +154,7 @@ func withRelease(cr *v1beta1.Release) helmClient.ArgsApplier {
 		config.SkipCRDs = cr.Spec.ForProvider.SkipCRDs
 		config.InsecureSkipTLSVerify = cr.Spec.ForProvider.InsecureSkipTLSVerify
 		config.PlainHTTP = cr.Spec.ForProvider.PlainHTTP
+		config.TakeOwnership = cr.Spec.ForProvider.TakeOwnership
 	}
 }
 
--- a/pkg/controller/cluster/release/release.go
+++ b/pkg/controller/cluster/release/release.go
@@ -154,7 +154,9 @@ func withRelease(cr *v1beta1.Release) helmClient.ArgsApplier {
 		config.SkipCRDs = cr.Spec.ForProvider.SkipCRDs
 		config.InsecureSkipTLSVerify = cr.Spec.ForProvider.InsecureSkipTLSVerify
 		config.PlainHTTP = cr.Spec.ForProvider.PlainHTTP
-		config.TakeOwnership = cr.Spec.ForProvider.TakeOwnership
+		// Only use TakeOwnership if requested AND not already used
+		// This prevents silent adoption of resources during upgrades after initial adoption
+		config.TakeOwnership = cr.Spec.ForProvider.TakeOwnership && !cr.Status.AtProvider.OwnershipTaken
 	}
 }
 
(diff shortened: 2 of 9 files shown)


pkg/controller/cluster/release/release.go
```diff
@@ -154,6 +154,7 @@ func withRelease(cr *v1beta1.Release) helmClient.ArgsApplier {
 		config.SkipCRDs = cr.Spec.ForProvider.SkipCRDs
 		config.InsecureSkipTLSVerify = cr.Spec.ForProvider.InsecureSkipTLSVerify
 		config.PlainHTTP = cr.Spec.ForProvider.PlainHTTP
+		config.TakeOwnership = cr.Spec.ForProvider.TakeOwnership
```
sergenyalcin: It seems that we need to add this statement to the namespaced controller.

apis/cluster/release/v1beta1/types.go
```diff
@@ -97,6 +103,9 @@ type ReleaseParameters struct {
 	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
 	// PlainHTTP uses insecure HTTP connections for the chart download
 	PlainHTTP bool `json:"plainHTTP,omitempty"`
+	// TakeOwnership ignore the check for helm annotations and take ownership
+	// of the existing resources.
+	TakeOwnership bool `json:"takeOwnership,omitempty"`
```
sergenyalcin, line 114: The feature makes sense, and `TakeOwnership` flows cleanly from the API field through `Args` into both `action.Install` and `action.Upgrade`. One thing I would push back on is making this a permanent spec field without any reset behavior.

Once a user sets `takeOwnership: true` to adopt a pre-existing release, that flag remains `true` on subsequent reconciles. On every Helm upgrade, instead of calling [`existingResourceConflict`](https://github.com/helm/helm/blob/v3.18.5/pkg/action/validate.go#L65), Helm calls [`requireAdoption`](https://github.com/helm/helm/blob/v3.18.5/pkg/action/validate.go#L41). The difference is that `existingResourceConflict` validates `meta.helm.sh/release-name` and `app.kubernetes.io/managed-by` annotations, while `requireAdoption` simply checks if the resource exists and takes it.

This only applies to resources new to the chart in a given upgrade (see [upgrade.go:339-351](https://github.com/helm/helm/blob/v3.18.5/pkg/action/upgrade.go#L339-L351), where `toBeCreated` is the diff between the old and new manifests). So the blast radius is narrower than "every resource, every time", but it's still dangerous.

An example: you set `takeOwnership: true` to migrate a manually-deployed nginx into the crossplane env. It works. You leave the flag set. After a while, a small `ConfigMap nginx-debug` is created in the same namespace for an investigation. Meanwhile, the nginx chart v1.1 ships, and it adds a resource named `nginx-debug`. On the next reconciliation, provider-helm upgrades the chart. Helm silently adopts and overwrites the new ConfigMap with the chart's version. No error, `Synced: True`. The manually created ConfigMap is gone, and the reconcile log shows nothing wrong.

Without `takeOwnership: true`, that same upgrade would have failed loudly: `UPGRADE FAILED: ConfigMap "nginx-debug" exists and cannot be imported into the current release`.

One design choice that came to mind (not thought through much) is treating adoption as a one-time event. After a successful install where `takeOwnership` was set, write `takeOwnership: false` back to the spec. The user declares intent to adopt; the controller executes it once and clears the flag. If the field docs frame this as a "run-once" migration tool rather than a permanent mode, it's safe. As a permanent mode, it quietly disables an important safety check.

sergenyalcin (commented):
Thanks, @stevendborrelli, I left two comments. One of them is about the design of the new API. Let's discuss it in the PR.

I focused on the `TakeOwnership` implementation (last 3 commits) because the other feature was developed in #324 and already had a review.

apis/cluster/release/v1beta1/types.go
bobh66: > After a successful install where takeOwnership was set, write takeOwnership: false back to the spec.

This can cause issues with controllers that might be reconciling the `Release`, whether it's a Crossplane composite or Argo/Flux/etc.

Maybe set a status field when the `takeOwnership` processing has completed, and future reconciles can use the status field to determine which code path to select.

sergenyalcin: Hmm, @bobh66, yes, this is better than backing to `false`. I wanted to write about the current API to see if we can find a better design option.

bobh66: This reminds me of the issue with the `Create` management policy importing an existing resource by default, and the need to be able to distinguish the scenario where the existing resource should _not_ be imported, and it should be an error case when the resource already exists and Crossplane wants to "create" it.

I originally thought of that as a `mustCreate` case, but maybe it's better represented as something like: 

```
managementPolicyConfig:
  create:
    importExisting: true
``` 

which in this case would translate to `takeOwnership: true`

The default for `importExisting` would need to be `true` since that's the current behavior in most/all providers, which may not be desirable in this scenario.  This is a little more nuanced since I think it will already import an existing Helm chart that has the same external-name, and this is extending that to ignore the annotation checks.

stevendborrelli: I've updated the implementation so that the state of takeOwnership is reflected in the status:

```go
// Only use TakeOwnership if requested AND not already used
		// This prevents silent adoption of resources during upgrades after initial adoption
		config.TakeOwnership = cr.Spec.ForProvider.TakeOwnership && !cr.Status.AtProvider.OwnershipTaken
```

sergenyalcin:
This PR is on top of #324. So, after the #324 merged and this branch is rebased, I will re-review this.

pkg/controller/cluster/release/release.go
```diff
@@ -154,6 +154,9 @@ func withRelease(cr *v1beta1.Release) helmClient.ArgsApplier {
 		config.SkipCRDs = cr.Spec.ForProvider.SkipCRDs
 		config.InsecureSkipTLSVerify = cr.Spec.ForProvider.InsecureSkipTLSVerify
 		config.PlainHTTP = cr.Spec.ForProvider.PlainHTTP
+		// Only use TakeOwnership if requested AND not already used
+		// This prevents silent adoption of resources during upgrades after initial adoption
+		config.TakeOwnership = cr.Spec.ForProvider.TakeOwnership && !cr.Status.AtProvider.OwnershipTaken
```
erhancagirici, line 159: few notes on this approach:
Take ownership is an imperative operation. We cannot "observe" it from the external object itself, in this case the Helm Release.

`cr.Status.AtProvider.OwnershipTaken` can only set by the `deploy()` in Create/Update calls. If for some reason this fails , the subsequent `Observe` calls cannot "rehydrate" `OwnershipTaken` status, so it will stay `false` until the next successful C/U call correctly persists it. As a result, the subsequent C/U operations will run with `config.TakeOwnership = true` because this condition will be satisfied.

The other edge case is:
After a successful take ownership operation, and setting `OwnershipTaken = true` in status, it will stay like that. If ,for some reason, you intend another take ownership operation further in the release's lifecycle, that won't be honored. You would need to switch the `spec.TakeOwnership` to False first, to clear "OwnershipTaken", then switch to True again to trigger the take ownership.

erhancagirici (approved):
Thanks for the PR! LGTM with notes on the potential caveats on the operational aspect. These do not affect the feature itself, just wanted to note.

Issue #283, opened 2025-04-26
Support helm's skip schema validation feature


markphillips100:
### What problem are you facing?
I'm trying to install a helm package but the's own schema is incorrect.  It's missing a property definition that the chart supports, but the implicit schema validation helm performs means I cannot specify the property.  I intent to raise the issue with the chart author but it would be nice to just skip validation for when these situations occur

### How could Crossplane help solve your problem?
Support helm's `--skip-schema-validation` switch.

PR #332, opened 2026-05-08
build: bump Helm SDK to latest v3 release (fixes JSON Schema HTTP $ref regression)

patrickleet:
## Summary

Bumps `helm.sh/helm/v3` from `v3.18.5` to `v3.20.2`. Single-line `go.mod` change plus the resulting `go mod tidy` churn — no source code changes in provider-helm itself.

## Why

Helm `v3.18.5` introduced a regression where `values.schema.json` files referencing remote `$ref` URLs over HTTPS fail at chart-load time with:

> `failing loading "https://raw.githubusercontent.com/...": invalid file url`

This is tracked by [helm/helm#31136](https://github.com/helm/helm/issues/31136). The fix — adding an `HTTPURLLoader` in `pkg/chartutil/jsonschema.go` — landed in [helm/helm#31138](https://github.com/helm/helm/pull/31138) and shipped in `v3.20.0` (Jan 2026) and `v3.20.2` (Apr 2026).

Because provider-helm consumes `chartutil` directly via `helm.sh/helm/v3/pkg/action` → `Install.Run()` → `chartutil.ValidateAgainstSchema`, *all* charts whose schema uses HTTPS `$ref` are currently uninstallable through provider-helm. Affected examples:

- [`zitadel/zitadel-charts` after their #532](https://github.com/zitadel/zitadel-charts/pull/532) — uses `https://raw.githubusercontent.com/yannh/kubernetes-json-schema/...` for ~20 fields including `affinity`, `env`, `extraContainers`, `ingress.tls`, `resources`, etc.
- Any chart that follows the same pattern of pulling Kubernetes type definitions from `yannh/kubernetes-json-schema`.

There is also an existing open feature request — [#283 *Support helm's skip schema validation feature*](https://github.com/crossplane-contrib/provider-helm/issues/283) — that would address a related concern (broken chart schemas) via the orthogonal `--skip-schema-validation` route. This PR does not implement that feature, but the SDK bump gives users a fixed schema validator out of the box.

## Verification

- `go build ./...` clean
- `go test ./...` — all tests pass, no API breakage between Helm 3.18 and 3.20 surfaced in provider-helm's call sites
- `make build && make xpkg.build.provider-helm` clean
- Live test on a Crossplane v2 control plane:
  - Before bump (`v1.2.0`, Helm SDK `v3.18.5`): installing `zitadel/zitadel:9.34.1` fails with `cannot install release: ... failing loading "https://raw.githubusercontent.com/...": invalid file url`

## Compatibility notes

- Helm `v3.20.0` is a minor-version bump; no public-API breakage for consumers using `pkg/action`, `pkg/chart/loader`, `pkg/chartutil`. Provider-helm's existing call sites compile and run unchanged.
- `go mod tidy` pulled forward a number of indirect Kubernetes deps to `0.35.x` — these are transitive and don't affect the provider's wire-API compatibility.
- Skips Helm `v3.19.x`, lands directly on the latest `v3.20.x` patch (`3.20.2`, Apr 2026) so consumers also pick up the post-3.20 fixes.

## Refs

- helm/helm#31136 (regression)
- helm/helm#31138 (fix, merged Aug 2025, shipped v3.20.0)
- helm/helm#31161 (follow-up not required for this fix)


patrickleet:
Can someone review this and/or #333?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/crossplane-contrib/provider-helm
 
-go 1.24.13
+go 1.25.0
```
lsviben: maybe lets bump to 1.25.10?

go.mod
```diff
@@ -15,16 +15,16 @@ require (
...
-	k8s.io/apiextensions-apiserver v0.33.3
-	k8s.io/apimachinery v0.34.2
-	k8s.io/client-go v0.34.2
+	google.golang.org/grpc v1.72.2
+	helm.sh/helm/v3 v3.20.2
+	k8s.io/api v0.35.1
+	k8s.io/apiextensions-apiserver v0.35.1
+	k8s.io/apimachinery v0.35.1
+	k8s.io/client-go v0.35.1
 	sigs.k8s.io/controller-runtime v0.19.1
```
lsviben, line 24: If we are bumping k8s, did you check if we should also bump the controller-runtime?

lsviben (commented):
Hey thanks for the change and sorry for waiting 🙏 

Left some minor comments, otherwise seems good to me.

go.mod
patrickleet: I checked this. Bumping controller-runtime to the version aligned with the k8s 0.35.x line (`v0.23.x`) does not compile with the current `github.com/crossplane/crossplane-runtime/v2 v2.0.0-rc.1`.

The failure is from newer controller-runtime client interfaces requiring `Apply`, e.g. Crossplane runtime's wrapped status client no longer satisfies `client.SubResourceWriter`. I left controller-runtime at `v0.19.1`; updating it looks like it should be a separate Crossplane runtime bump.

go.mod
patrickleet: Updated this to `go 1.25.10` and set CI `GO_VERSION` to `1.25.10` as well.

patrickleet: I also bumped golangci-lint from `v2.1.2` to `v2.4.0`. The old linter binary was built with Go 1.24, so it fails once the module targets Go 1.25.

patrickleet:
Updated the PR to use Helm SDK `v3.21.0` instead of `v3.20.2`, since `v3.21.0` is now the latest Helm v3 release.

I also added a regression test for the actual failure mode: Helm install with a `values.schema.json` that resolves an HTTP `$ref`. Forced against Helm `v3.18.5` with a temporary modfile, the test fails with `invalid file url`; with Helm `v3.21.0`, it passes.

go.mod
patrickleet: I also tested moving this PR all the way to the latest Crossplane runtime, `github.com/crossplane/crossplane-runtime/v2 v2.3.1`.

Provider Helm itself can be made to build/test against it, but it is not a clean dependency bump right now. Runtime v2.3.1 no longer contains the old `crossplane-runtime/v2/apis/common*` packages, so the provider code needs to move API imports to `github.com/crossplane/crossplane/apis/v2/core/v2` and update embedded types like `ResourceSpec` -> `ClusterManagedResourceSpec` and `ResourceStatus` -> `ManagedResourceStatus`.

The larger blocker is `github.com/crossplane-contrib/provider-kubernetes`. The latest released version is currently `v1.2.1`, and it still imports the removed `crossplane-runtime/v2/apis/common*` packages. 

The Crossplane runtime bump should wait for a compatible provider-kubernetes release.

patrickleet: Here's a PR for that https://github.com/crossplane-contrib/provider-kubernetes/pull/480 but as I said, it's a separate issue. I'll make a follow up PR for provider-helm once these are merged.

lsviben: Got it, thanks for the follow-ups!

lsviben (approved):
Thanks @patrickleet !

Issue #329, opened 2026-04-08
Provider-helm considers a synced/unready release as uptodate


Kidswiss:
### What happened?

We often end up with releases that will be synced, but not ready. If that happens it will take the provider `pollInterval` (10 minutes) to actually mark the release as synced/ready.

So it will take 10 minutes for the claim to get ready and the `connectionDetails` to be written, even though everything is already deployed.

The logic in the provider only considers the synced condition and the rollbacks for the release to be `ResourceUpToDate` as seen here: https://github.com/crossplane-contrib/provider-helm/blob/main/pkg/controller/cluster/release/release.go#L258. Is there a reason that the readiness is not considered as well?

### How can we reproduce it?

Unfortunately it's a bit of a race condition, so I don't have exact steps to reproduce. It usually happens if the release depends on other resources provisioned by the composition, like secrets. In which case the release will go to unsynced/unready with state `failed` and properly retry every 30 seconds until the dependency exists. Then the release will go to synced/unready with state `deployed` and be stuck for 10 minutes.

Restarting the provider or annotating the release will instantly trigger a new reconcile and fix the conditions.

### Further observations

Looking at `IsUpToDate`: https://github.com/crossplane-contrib/provider-helm/blob/main/pkg/controller/cluster/release/observe.go#L64

It seems to only check the spec and not the release's state. So there might be an edge case where `AtProvider.State=failed` while `status.Synced=true`. But maybe I'm missing something.

### What environment did it happen in?
Crossplane version: any

PR #333, opened 2026-05-13
Update helm sdk to v4

### LINKED ISSUE #214: Release Ready status should be aggregation of readines of object deployed within a release

Ready status of a Helm release becomes equal to `true` as soon as Helm creates objects that are a part of the given release. However, this does not necessary means that the workload is really ready, because images need to be pulled from registries, pods scheduled and started, etc.

In a complex compositions, quite often deployment of another chart or objects depends on readiness of their dependencies.
Semantically would make sense to mark a Helm release ready when the enclosed resources are really ready. Making the equivalence with being deployed is imprecise.

Even today, the initial readiness could be emulated by setting `.spec.forProvider.wait` to `true`. In that case, the provider is going to block and wait until Helm checks and returns back when all resources become ready.

Given that `wait:true` is blocking the reconcile loop, it has impact on other releases sitting in the queue and waiting to be reconciled. Also, once readiness is detected, we do not try to perform any future checks - a Helm release remains ready even if the workload crashes and does not recover at all.

Helm release should become ready only when all deployed objects part of the release becomes ready. This check should be performed on every release reconciliation and ready condition should be updated accordingly.

### LINKED ISSUE #314: Upgrade Helm to version 4
With Helm 4 being released a while ago with lots of goodies, https://helm.sh/docs/overview/.
One of the main features (to me) is the switch to server-side apply, avoiding some of the previously seen side-effects (e.g. https://github.com/kubernetes/kubernetes/issues/135423). Also there is a new Chart format (v3) planned, support for this would require Helm 4.

Switch to Helm 4 in the Helm provider!

1. Helm themselves calls for SDK users to use their new stable API and report back if there are any issues, see https://helm.sh/docs/overview/#high-priority.
2. With Helm 4 being mostly a drop in replacement, it's very likely that Helm 3 will not receive updates mid or even long-term, see https://helm.sh/docs/topics/version_skew/#supported-version-skew.

jonasz-lasut:
Fixes #314 Fixes #214

Updates helm sdk to v4. Notable changes:
* Server-side apply is now a default for Helm, however old release that used client-side apply will continue to do so. We could override this behavior to always prefer SSA with forced conflict resolution by settings `ServerSideApply` and `ForceConflicts` on helm  `actions` to `true`. I'll happily get a feedback if we want to do so but I think that default behavior is fine. Maybe it could be introduced behind a spec field e.g. `ssaForceConflicts` or similar
* Simple boolean `wait` has been replaced by wait strategies
CLI deprecation shim at pkg/cmd/flags.go:88-93 in helm v4 explicitly maps the legacy booleans:
wait=true → kube.StatusWatcherStrategy
wait=false → kube.HookOnlyStrategy
I've mapped them accordingly in code, wait true now waits based on KStatus
* Replaced in-tree `LegacyTracker` and `ModernTracker` with their implementation from `crossplane-runtime` as they were merged there with a PR from Erhan some time ago
* Update `go` and `golangci-lint` to satisfy the requirements of helm sdk v4
* Added linter exemptions for legacy types and deprecated `controller-runtime` events API with comments
* Added content cache dir, helm now properly caches charts based on content as per https://github.com/helm/helm/pull/31165

I have:

Run E2E tests locally


pkg/clients/helm/client.go
```diff
@@ -99,9 +101,7 @@ func NewClient(log logging.Logger, restConfig *rest.Config, argAppliers ...ArgsA
 
 	actionConfig := new(action.Configuration)
 	// Always store helm state in the same cluster/namespace where chart is deployed
-	if err := actionConfig.Init(rg, args.Namespace, helmDriverSecret, func(format string, v ...interface{}) {
-		log.Debug(fmt.Sprintf(format, v...))
-	}); err != nil {
+	if err := actionConfig.Init(rg, args.Namespace, helmDriverSecret); err != nil {
 		return nil, err
```
sergenyalcin: Now we remove this logging. Although, the logging formatter function is removed, should we still log something here?

pkg/clients/helm/client.go
```diff
@@ -415,12 +434,20 @@ func (hc *client) PullAndLoadChart(mg resource.Managed, creds *RepoCreds) (*char
 	return chart, nil
 }
 
-func (hc *client) GetLastRelease(release string) (*release.Release, error) {
-	return hc.getClient.Run(release)
+func (hc *client) GetLastRelease(name string) (*release.Release, error) {
+	r, err := hc.getClient.Run(name)
+	if err != nil {
+		return nil, err
```
sergenyalcin, line 455: There are some unwrapped errors. Could you please wrap them?

pkg/clients/helm/client.go
```diff
@@ -415,12 +434,20 @@ func (hc *client) PullAndLoadChart(mg resource.Managed, creds *RepoCreds) (*char
 	return chart, nil
 }
 
-func (hc *client) GetLastRelease(release string) (*release.Release, error) {
-	return hc.getClient.Run(release)
+func (hc *client) GetLastRelease(name string) (*release.Release, error) {
+	r, err := hc.getClient.Run(name)
```
sergenyalcin, line 453: Just out of curiosity, as far as I understand, instead of directly returning the result of Run, we started to make sure the type of the returned object. Do we have a scenario that Run will return non Release object?

apis/cluster/release/v1alpha1/register.go
```diff
@@ -45,6 +49,11 @@ var (
 	ReleaseGroupVersionKind = SchemeGroupVersion.WithKind(ReleaseKind)
 )
 
-func init() {
```
sergenyalcin, line 48: nit: In the other official provider repositories we are using the old approach. Just out of curiosity, do we have a reason to switch this version of schema registration?

pkg/clients/helm/client.go
```diff
@@ -121,37 +121,56 @@ func NewClient(log logging.Logger, restConfig *rest.Config, argAppliers ...ArgsA
...
+	}
+	pc.InsecureSkipTLSVerify = args.InsecureSkipTLSVerify
 	pc.PlainHTTP = args.PlainHTTP
 
 	gc := action.NewGet(actionConfig)
 
+	waitStrategy := kube.HookOnlyStrategy
+	if args.Wait {
+		waitStrategy = kube.StatusWatcherStrategy
+	}
```
sergenyalcin, line 159: Question: Do we have a breaking behavioral change? Does the new parameters change the older wait behavior? If yes what about mentioning them in here as a code comment.

sergenyalcin (commented):
Thanks @jonasz-lasut for this important v4 bump. I've left a few comments for your consideration.

apis/cluster/release/v1alpha1/register.go
jonasz-lasut: it was forced by the update of controller-runtime and k8s modules that came with updating helm to v4. old approach is now deprecated and was getting flagged by golangci-lint

pkg/clients/helm/client.go
jonasz-lasut: Not today — the assertion is forced by the new signatures. Helm v4 changed the action `Run` methods to return `release.Releaser` instead of `*release.Release`, and `Releaser` is literally an alias for `any`, an intentional forward-compat hook for future release schema versions:

https://github.com/helm/helm/blob/v4.2.3/pkg/release/interfaces.go#L25

In practice the v1 action package always returns `*release/v1.Release`, so the `!ok` branch is defensive-only, but the assertion is the only way to get the concrete type back out of `any` here.

pkg/clients/helm/client.go
jonasz-lasut: Yes — for `wait: true` there is a behavioral change, and we want to keep it. Helm v4 ships three strategies (https://github.com/helm/helm/blob/v4.2.3/pkg/kube/client.go#L105-L119):

* `StatusWatcherStrategy` — event-driven, waits on kstatus readiness
* `LegacyStrategy` — Helm 3's periodic poller, kept for compatibility
* `HookOnlyStrategy` — waits for hook Pods/Jobs only (v3's `wait: false` behavior)

The mapping here follows helm's own shim for the deprecated `--wait` flag (https://github.com/helm/helm/blob/v4.2.3/pkg/cmd/flags.go#L87-L94): `wait: false` → `HookOnlyStrategy`, unchanged vs v3; `wait: true` → `StatusWatcherStrategy`, kstatus instead of the legacy poller.

Keeping the watcher instead of pinning `LegacyStrategy` is deliberate — the legacy poller has known false positives. **Tested (provider-helm v1.3.1 / helm v3, kind):** with `replicas: 1` and `rollingUpdate.maxUnavailable: 1` (podinfo's chart default) the required ready count is `replicas - maxUnavailable = 0`, so upgrading to a broken image reports "deployed" and the Release goes Synced+Ready with every pod in ImagePullBackOff — one mechanism behind #329. kstatus waits for actual readiness, so the visible change is that some releases which previously went Ready instantly now wait until resources are actually ready (or hit the configured timeout).

I've added a code comment at this mapping documenting the change; it will be pushed together with the changes for the remaining comments.

pkg/clients/helm/client.go
jonasz-lasut: Good catch and it turned out to be worse than just dropping the bridge: with no handler set, helm v4's `Configuration` **discards** its internal logs entirely (`LogHolder.Logger()` falls back to `slog.DiscardHandler`, https://github.com/helm/helm/blob/v4.2.3/internal/logging/logging.go#L108-L113), so e.g. the kstatus wait diagnostics were silently dropped not even reaching stderr.
Added a small slog handler

pkg/clients/helm/client.go
jonasz-lasut: Added wrapped errors

sergenyalcin (approved):
Thanks @jonasz-lasut LGTM!

PR #280, opened 2025-03-11
Add support for revisions history limit

hudymi:
### Description of your changes

I added support for the max history limit configuration. Otherwise, we have lots of big secrets on the cluster, because of hardcoded 20 revisions history. For backward compatibility, I left the 20 history limit by default (Helm cli has 10 as the default).

https://helm.sh/docs/helm/helm_upgrade/#options

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.

### How has this code been tested

Tested on live cluster and upgraded release multiple times.

[contribution process]: https://git.io/fj2m9

PR #338, opened 2026-06-19
Add support for configurable revision history limit

portswigger-amason:

This adds a `maxHistory` field to the `Release` API that limits the number of revisions saved per release, threaded through to the Helm upgrade client. Previously the history limit was hardcoded to 20, which can accumulate a large number of release-history `Secret`s on clusters with many releases. It defaults to `20` to preserve the existing behaviour; set `0` for no limit.

This recreates and rebases @hudymi's original PR #280 onto the current Crossplane v2 layout — applying it to both the cluster (`helm.crossplane.io`) and namespaced (`helm.m.crossplane.io`) `Release` APIs, and using an idiomatic optional `json:"maxHistory,omitempty"` field. Full credit to @hudymi for the original change; opening a fresh PR since the original needed a rebase (per the maintainer's request on #280).

I have:

- [x] Added or updated unit tests where applicable.

- `go build ./...`, `go vet`, and the existing unit tests pass for `pkg/clients/helm` and both the cluster and namespaced release controllers.
- CRDs regenerated via `go generate` (controller-gen); `maxHistory` appears in both `Release` CRDs as an optional integer defaulting to `20`.

[contribution process]: https://github.com/crossplane/crossplane/blob/main/CONTRIBUTING.md

### DIFF
--- a/apis/namespaced/release/v1beta1/types.go
+++ b/apis/namespaced/release/v1beta1/types.go
@@ -108,6 +108,10 @@ type ReleaseParameters struct {
 	// This prevents silent adoption of unrelated resources during chart upgrades.
 	// Use this field to migrate manually-deployed Helm releases into Crossplane management.
 	TakeOwnership bool `json:"takeOwnership,omitempty"`
+	// MaxHistory limits the maximum number of revisions saved per release. Use 0 for no limit.
+	// +optional
+	// +kubebuilder:default:=20
+	MaxHistory int `json:"maxHistory,omitempty"`
 }
 
 // ReleaseObservation are the observable fields of a Release.
--- a/apis/cluster/release/v1beta1/types.go
+++ b/apis/cluster/release/v1beta1/types.go
@@ -112,6 +112,10 @@ type ReleaseParameters struct {
 	// This prevents silent adoption of unrelated resources during chart upgrades.
 	// Use this field to migrate manually-deployed Helm releases into Crossplane management.
 	TakeOwnership bool `json:"takeOwnership,omitempty"`
+	// MaxHistory limits the maximum number of revisions saved per release. Use 0 for no limit.
+	// +optional
+	// +kubebuilder:default:=20
+	MaxHistory int `json:"maxHistory,omitempty"`
 }
 
 // ReleaseObservation are the observable fields of a Release.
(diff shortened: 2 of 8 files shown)


portswigger-amason:
Please can I get an approval on this?

portswigger-amason:
I've rebased this onto the latest `main` and resolved the merge conflict — the PR is now mergeable again. Would appreciate a review when you have a moment 🙏 @turkenh @lsviben @erhancagirici @sergenyalcin @turkenf @ulucinar

apis/namespaced/release/v1beta1/types.go
```diff
@@ -102,6 +102,10 @@ type ReleaseParameters struct {
 	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
 	// PlainHTTP uses insecure HTTP connections for the chart download
 	PlainHTTP bool `json:"plainHTTP,omitempty"`
+	// MaxHistory limits the maximum number of revisions saved per release. Use 0 for no limit.
+	// +optional
+	// +kubebuilder:default:=20
```
erhancagirici, line 113: one small concern regarding the defaulting:
This is indeed backward-compatible in terms of functionality. However after upgrading the provider, existing MRs persist this into their spec. This could be a potential breaking API behavior for:
- In gitops scenarios, can potentially cause a spec diff
- This does not respect late-init management policies, gets persisted regardless

IMHO, for full backwards-compatibility, omitting `spec.forProvider.maxHistory` should not modify spec (possibly using a pointer field), and runtime should process only when value is explicitly specified. Omitting should keep today's behavior (hard-coded 20 revisions), and for no-limit, consumer should explicitly set `0`.

erhancagirici (commented):
@portswigger-amason thanks for the PR! I've left a small comment for consideration.

erhancagirici (approved):
Giving [my previous comment](https://github.com/crossplane-contrib/provider-helm/pull/338/changes#r3499032776) a second thought, I think the defaulting is acceptable. I've rebased the PR again (there was a new conflict with latest merges).
Thanks again for the contribution @portswigger-amason 🎉 LGTM

portswigger-amason:
@erhancagirici Thank you for the approval. Would you be able to merge for me?

PR #371, opened 2026-08-05
fix(release): skip set if optional source missing

### LINKED ISSUE #265: Skip set if valueFrom source ref is optional and missing

With a release spec like this:
```yaml
forProvider:
  set:
    - name: foo
      valueFrom:
        secretKeyRef:
          name: bar
          namespace: baz
          key: optionalKey
          optional: true
```

The release installation fails with `missing value for --set` if the secret is missing.

To make optional source refs useful in the context of set would it not make sense to also skip the set if the value is missing? Two possible solutions I see would be to either check if the value source is optional and skip it or introduce a way of marking a set optional as well.

jonasz-lasut:

A `set` entry whose `valueFrom` points at an `optional: true` secret/configmap ref still failed the whole values composition with `missing value for --set` when the source (or its key) was missing — making `optional` meaningless for `set`.

`getDataValueFromSource` already resolves optional+missing sources to an empty value; `composeValuesFromSpec` now skips such `set` entries instead of erroring. Behavior is unchanged for non-optional refs (still an error) and for a `set` with neither `value` nor `valueFrom` (still `missing value for --set`). Applied to both the cluster-scoped and namespaced controllers.

Fixes #265

I have:

New table-test cases in both scopes — optional secret missing, optional key missing, optional configmap missing — all reproduce the `missing value for --set` failure on the previous code and pass with the fix. `make reviewable` (generate, lint, unit tests) is green.

[contribution process]: https://github.com/crossplane/crossplane/blob/main/contributing/README.md

### DIFF
--- a/pkg/controller/cluster/release/values.go
+++ b/pkg/controller/cluster/release/values.go
@@ -77,6 +77,9 @@ func composeValuesFromSpec(ctx context.Context, kube client.Client, spec v1beta1
 		}
 
 		if v == "" {
+			if optionalValueFrom(s.ValueFrom) {
+				continue
+			}
 			return nil, errors.New(errMissingValueForSet)
 		}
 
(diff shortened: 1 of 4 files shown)


sergenyalcin (approved):
Thanks @jonasz-lasut LGTM!

PR #385, opened 2026-08-18
feat(release): observe deploy facts via helm release labels

### LINKED ISSUE #382: Release digest/ownership status is not reliably observable, and chart URL drift goes undetected
### What happens

- `status.atProvider.digest` and `ownershipTaken` are set during Create, but crossplane-runtime reverts any status written during Create when it persists critical annotations (`pkg/reconciler/managed/reconciler.go`), so both are lost. `generateObservation` additionally zeroes `ownershipTaken` on every Observe, so the re-adoption guard only holds for one poll after each upgrade.
- `isUpToDate` never compares `spec.forProvider.chart.url`, so changing the chart URL on a URL-only deployment is never detected as drift.

### Proposed fix

Persist the deploy facts as helm release labels - helm stores them on the release Secret atomically with the deploy and returns them on read:

- `release.helm.crossplane.io/digest-hash`, `.../url-hash`, `.../ownership-taken`

Observe rehydrates `status.atProvider` from the labels (with a fallback for releases deployed before label support), and `isUpToDate` compares the `url-hash` label against the spec URL to detect URL drift.

### Environment

provider-helm `main` (post helm-v4), `helm.sh/helm/v4 v4.2.3`, `crossplane-runtime/v2 v2.3.3`.

jonasz-lasut:
Persist the facts a Release needs to observe about its own deploy as custom helm release labels, which helm stores on the release Secret atomically with the deploy and returns on GetLastRelease. This makes them genuinely observable, unlike status fields: crossplane-runtime resets status set during Create when it persists critical annotations, and generateObservation zeroed OwnershipTaken on every Observe, which broke the re-adoption guard one poll after each upgrade.

Labels, written on install and upgrade:
- release.helm.crossplane.io/digest-hash: sha256-<first 56 hex> of the effective chart digest (label values cap at 63 chars; the encoding is used for equality drift-detection only, pull-time verification keeps the full digest)
- release.helm.crossplane.io/url-hash: sha256-<56 hex> of sha256(url)
- release.helm.crossplane.io/ownership-taken: sticky true once takeOwnership was exercised, so later upgrades never silently re-adopt

Stale digest/url values are removed on upgrade via helm's null-value label deletion; Observe rehydrates status from the labels and falls back to the previously persisted status for releases deployed before label support.

isUpToDate now compares the url-hash label against the encoded spec URL, so URL-only deployments detect drift. URL mode stops late-initializing name and version (a stale late-initialized version would loop forever after the first URL change; the deployed version stays visible in status.atProvider.version), keys the name and version checks on URL presence, and reconciles OCI URL versions with the spec version via resolveEffectiveVersion, reporting conflicts as drift so the deploy surfaces its hard error.

E2E: url-drift examples whose post-assert hook changes the chart URL and requires the release to roll to the new chart version via the url-hash label, and digest-hash label plus status digest assertions on the existing digest examples (both cluster and namespaced scopes).

Fixes #382

I have:

Added E2E test for url-drift


pkg/controller/cluster/release/observe.go
```diff
@@ -67,6 +68,33 @@ func generateObservation(in *release.Release) v1beta1.ReleaseObservation {
...
+func rehydrateFromLabels(cr *v1beta1.Release, rel *release.Release, lastDigest string, lastOwnershipTaken bool) {
+	if v, ok := rel.Labels[helmClient.LabelDigestHash]; ok {
+		specDigest := helmClient.EffectiveChartDigest(cr.Spec.ForProvider.Chart.URL, cr.Spec.ForProvider.Chart.Digest)
+		if v == helmClient.EncodeDigestLabel(specDigest) {
+			// The deployed digest matches the spec: surface it in full
+			// fidelity, since the label only stores a truncated encoding.
+			cr.Status.AtProvider.Digest = specDigest
+		} else {
+			// Drifted: only the encoded form of the deployed digest is known.
+			cr.Status.AtProvider.Digest = v
```
erhancagirici: status.atProvider.digest can be permanently set to a non-digest

  Deploy repository-mode with digest: sha256:aaaa… (label written), then remove chart.digest from spec leaving version unchanged. In isUpToDate (180–190) both specDigestEnc and in.Chart.Digest are empty → neither case
  fires → up-to-date, so no upgrade ever clears the stale label. Every later Observe takes the mismatch branch and writes sha256-<first 56 hex> — dash-separated and truncated — into a field documented as "last
  successfully deployed chart digest" (CRD pattern ^sha256:[a-f0-9]{64}$). It also feeds back as lastDigest on subsequent Observes, and anything consuming it (composition patches, users copying it into a spec) gets
  garbage.

pkg/controller/cluster/release/observe.go
```diff
@@ -119,15 +150,46 @@ func isUpToDate(ctx context.Context, kube client.Client, spec *v1beta1.ReleaseSp
 		return true, nil
 	}
 
-	// Check version match only if version is specified in spec
-	// For digest-only deployments, skip version check as version is optional
-	if in.Chart.Version != "" && in.Chart.Version != ocm.Version && in.Chart.Version != devel {
-		return false, nil
+	if in.Chart.URL == "" {
```
erhancagirici: No escape hatch behind a floating tag

  With url: `oci://…/podinfo:latest` plus `version: 6.10.2`, bumping version used to force a re-pull. Now nothing compares versions, the URL string is unchanged so url-hash matches, and no digest is pinned → the release
  sticks to the old chart with no way to force a re-pull short of editing an unrelated field. The loop-avoidance motivation for v-prefixed/:latest tags is legitimate; worth documenting or a devel-style opt-in.

pkg/controller/cluster/release/release.go
```diff
@@ -163,6 +163,7 @@ func withRelease(cr *v1beta1.Release) helmClient.ArgsApplier {
 		config.TakeOwnership = cr.Spec.ForProvider.TakeOwnership && !cr.Status.AtProvider.OwnershipTaken
```
erhancagirici: Nit, double-check for Update, not sure if the Create path is relevant for TakeOwnership. In any case, I believe we cannot get past this with current crossplane-runtime impl, since status updates at Create cannot be persisted in the same reconcile.

TakeOwnership guard is one reconcile stale:

  `withRelease` derives TakeOwnership and the sticky label from `cr.Status.AtProvider.OwnershipTaken` at Connect time, but the label is only read back in `Observe`, after the client is built. On the reconcile right after
  Create (where runtime has just reset status), `Connect `still computes TakeOwnership=true; if that reconcile finds drift and calls Update, helm re-adopts despite ownership-taken already being set. Closing it needs the
  flag applied at deploy time, or the label read before client construction.

pkg/controller/cluster/release/observe.go
```diff
@@ -119,15 +150,46 @@ func isUpToDate(ctx context.Context, kube client.Client, spec *v1beta1.ReleaseSp
...
+		}
 	}
 
-	// Check if digest has changed - if specified, compare against last synced digest
-	// Note: Chart metadata doesn't include OCI digest, so we store it in status
-	if in.Chart.Digest != "" && s.AtProvider.Digest != "" && in.Chart.Digest != s.AtProvider.Digest {
+	// Digest drift: prefer the label written at deploy time, which makes the
+	// deployed digest observable. Releases deployed before label support fall
+	// back to the digest persisted in status.
+	specDigestEnc := helmClient.EncodeDigestLabel(helmClient.EffectiveChartDigest(in.Chart.URL, in.Chart.Digest))
```
erhancagirici, line 188: nit, can be a follow-up:

Adding a digest to an already-deployed release is still a no-op:

  A release the new provider deployed without a digest has no digest-hash label (written as null), so the legacy else if branch applies and requires s.AtProvider.Digest != "" → no drift. Setting chart.digest: sha256:X
  afterward never re-pulls. The label mechanism now makes "label absent" a positive signal, so this is detectable if you distinguish provider-written from pre-label releases.

pkg/controller/cluster/release/observe.go
```diff
@@ -119,15 +150,46 @@ func isUpToDate(ctx context.Context, kube client.Client, spec *v1beta1.ReleaseSp
...
+	if deployedDigest, ok := observed.Labels[helmClient.LabelDigestHash]; ok {
```
erhancagirici, line 189: nit, can be follow-up, Unpinning a digest is a silent no-op:

 In the labelled branch, when the spec drops its digest, specDigestEnc == "" and in.Chart.Digest == "", so neither switch case fires →
  up-to-date. Nothing redeploys, the stale digest-hash label is never cleaned up (only an upgrade would rewrite it, and no upgrade happens), and rehydrateFromLabels keeps reporting the truncated sha256-<A[:56]> as
  status.atProvider.digest forever.

pkg/controller/cluster/release/observe.go
erhancagirici: possible nit:

With the name check and version check now gated on `in.Chart.URL == ""`, `url-hash` is the only URL-drift signal. But for a repository-mode deploy `releaseLabels` encodes "no URL" as `"null"`,
  which `withoutDeletedLabels` strips on install and `mergeCustomLabels` deletes on upgrade — so the key is **absent**, indistinguishable from a pre-label release.

  Migrating an existing Release to OCI (`chart.url: oci://…/podinfo:6.10.2` + matching `chart.version` bump) therefore hits: name check skipped, version check skipped, `ok == false` here, no digest → `isUpToDate`
  returns true. The release keeps serving the repository chart indefinitely, `Synced`/`Available`. The version bump used to force this upgrade pre-PR, so this is a regression rather than a pre-existing gap.

  Note the asymmetry: the reverse direction already works — `NotUpToDate_URLClearedSwitchesToRepoMode` passes precisely because the label is *present* there.

  Fix is the encoding, not this check: have `labelValueOrDelete` write `""` instead of `LabelValueDelete` for these two keys. Empty string is a valid label value and helm only treats the literal `"null"` as deletion,
  so `""` persists and this comparison reads `"" != <new URL hash>` → drift, mirroring the URL→repo case. Genuine legacy releases stay absent and keep their documented behavior. The same change closes the digest-add
  and digest-unpin no-ops

erhancagirici (commented):
thanks @jonasz-lasut LGTM in general, left a few AI-assisted nit comments, AFAIU they are mostly edge-cases and raised due to us keeping some old (non-ideal) behavior for the sake of no behavior change for those.
 
We can follow up on those, after you also provide your opinion their relevancy.

pkg/controller/cluster/release/observe.go
jonasz-lasut: good catch the root cause was the unpin no-op you flagged, I've added the following to fix it
unpinning now drifts and redeploys, which rewrites the label as "", and rehydrateFromLabels no longer writes the encoded form into status at all. On a label mismatch it keeps the last persisted full digest, which is the digest of the last successful deploy (or empty right after Create, when the runtime has reverted status). status.atProvider.digest now only ever holds a full sha256:<64 hex> or is empty.

pkg/controller/cluster/release/release.go
jonasz-lasut: another good catch and it was actually a serious regression
deploy() also sets status.patchesSha, which the same post-Create revert clears, so on the next reconcile hasUpdates compares the real SHA against "" and forces an Update for every Release with patchesFrom. That Update ran with TakeOwnership=true.

fixed in f8e4d1c: TakeOwnership and the release labels moved out of the Connect-time Args into a DeployOptions value passed to Install/Upgrade per call

pkg/controller/cluster/release/observe.go
jonasz-lasut: The "" encoding alone didn't close this one: with the spec digest empty, specDigestEnc == "" disabled the compare, so a present sha256-<A> label was never looked at. The compare is now unconditional whenever the label is present, so unpinning drifts, redeploys and rewrites the label as "" (NotUpToDate_DigestUnpinned). One behavior change to be aware of: unpinning now costs a helm revision, where before (also on main) it was a silent no-op.

pkg/controller/cluster/release/observe.go
jonasz-lasut: fixed now as it was a small change to implement

the digest-hash label is now written as "" when the spec has no digest, so a release this provider deployed unpinned has the label present and the existing compare reports drift when a digest is added later (NotUpToDate_DigestAddedToUnpinnedRelease). Releases that predate label support keep the old status-based check, so for them this stays as it was on main.

pkg/controller/cluster/release/observe.go
jonasz-lasut: The "null" marker collapsed "deployed without a URL" into "deployed before labels", and "" is a valid label value that helm stores and merges verbatim, so both labels are now written that way; LabelValueDelete and the install-time strip are gone (3a14185, NotUpToDate_RepoReleaseMigratedToURL).

Thanks for catching that as it was not a small nit but a gap

pkg/controller/cluster/release/release.go
```diff
@@ -355,7 +387,7 @@ func (e *helmExternal) deploy(ctx context.Context, cr *v1beta1.Release, action d
 		}
 	}
 
-	rel, err := action(meta.GetExternalName(cr), chart, cv, p)
+	rel, err := action(meta.GetExternalName(cr), chart, cv, p, deployOptions(cr))
```
erhancagirici: nit, an edge case, the triggering case is somewhat rare but for future general correctness:

In late init scenarios, the `needsUpdate` block above does a `kube.Update` on spec.
This clobbers the `status` fields we set previously at `Observe` in the same reconciliation (not persisted yet, those are still in-memory).
`deployOptions(cr)` now reads from the clobbered status here.
One quick fix would be to calculate  `deployOptions(cr)` before the late init, then pass it here.

AFAIU this wasn't particularly an issue before this PR, because nothing depended on the `status` in `action()` parameters, at this point.


Triggering scenario: repository-mode release adopted via takeOwnership on Create; next reconcile rehydrates OwnershipTaken=true, GitOps has stripped the late-initialized chart.version, another field drifts → Update → late-init
  Fix: hoist opts := deployOptions(cr) above the late-init block. It only depends on Chart.URL/Chart.Digest and the status, none of which late-init touches.
```

Note: the late-init status clobber still there. We are not affected by it because, we set each status field that `Observe()` adjusted are set again at `deploy()`, so no status missed.

File: pkg/controller/cluster/release/release.go [hunk 7]
```diff
@@ -178,6 +169,38 @@ func withCABundle(caBundle []byte) helmClient.ArgsApplier {
...
+// status.atProvider from the release labels on this same object, so this sees
+// that ownership was already taken even on the reconcile right after Create,
+// when crossplane-runtime has reverted the status written there.
+func deployOptions(cr *v1beta1.Release) helmClient.DeployOptions {
+	// Only take ownership if requested AND not already taken. This prevents
+	// silent adoption of resources during upgrades after the initial adoption.
+	takeOwnership := cr.Spec.ForProvider.TakeOwnership && !cr.Status.AtProvider.OwnershipTaken
+	return helmClient.DeployOptions{
+		TakeOwnership: takeOwnership,
+		Labels:        releaseLabels(cr.Spec.ForProvider.Chart, takeOwnership),
```
erhancagirici: very nit, I don't think this is worth implementing but noting it for recording purposes. In fact, can be considered a choice of behavior.

the `takeOwnership` label (not the option itself) is not back-filled.

`releaseLabels()` adds `LabelOwnershipTaken` only when the effective `takeOwnership` is true.

Example scenario is: in a pre-PR version of the provider with takeOwnership feature, suppose you adopted a release and completed the take ownership operation. i.e. MR has `status.atProvider.ownershipTaken = true` now.

Now with this post-PR provider, when rehydrating:
The effective `takeOwnership` is `false` (correct)
`DeployOption.TakeOwnership` becomes `false` (correct, we don't want that action)
The label is unmodified, so `ownershipTaken` stays MR status-only.

This works as expected during MRs lifetime.
Only edge is: Deleting and re-importing the MR with same external name, would lose the status (the only information), therefore would execute "take ownership" action in the reconcile.

Depending on the situation, this can also be considered desired. Just flagging the behavior here.

Summary is, the deploy option of take ownership itself and writing the label might be distinguished, e.g
write the label when `cr.Status.AtProvider.OwnershipTaken || takeOwnership`

pkg/controller/cluster/release/release.go
```diff
@@ -289,7 +314,10 @@ func (e *helmExternal) Observe(ctx context.Context, mg resource.Managed) (manage
 		if err != nil {
 			return managed.ExternalObservation{}, errors.Wrap(err, "cannot get connection details")
 		}
-		if cr.Status.AtProvider.Digest == "" {
+		// Legacy releases without a digest label: assume the deployed digest
+		// matches the spec once deployed and synced. Labeled releases carry
+		// the deployed digest themselves and never need this.
+		if _, ok := rel.Labels[helmClient.LabelDigestHash]; !ok && cr.Status.AtProvider.Digest == "" {
```
erhancagirici: nit, This needs double-checking, AFAIU probably valid also pre-PR, can be follow-up:

Digest pinning a pre-PR release is a permanent no-op, and status back-fills a unverified digest.

On a label-less release, `isUpToDate` falls to none of the conditions;  so a release deployed without a digest has "", then adding chart.digest isn't drift. `Observe` then
  reaches here and writes `cr.Status.AtProvider.Digest = cr.Spec.ForProvider.Chart.Digest`,  asserting a digest nothing verified. 
In the next reconcile `spec == status`, so the pin is locked in as satisfied, no upgrade is  triggered, so the release never gains labels and never escapes the legacy path. 
Removing a digest from a legacy release is likewise inert (branch gated on in.Chart.Digest != "").

pkg/controller/cluster/release/observe.go
```diff
@@ -119,15 +150,54 @@ func isUpToDate(ctx context.Context, kube client.Client, spec *v1beta1.ReleaseSp
...
 
-	// Check version match only if version is specified in spec
-	// For digest-only deployments, skip version check as version is optional
-	if in.Chart.Version != "" && in.Chart.Version != ocm.Version && in.Chart.Version != devel {
+	// In URL mode the deployed chart's metadata version is deliberately not
+	// compared: an OCI URL tag is an arbitrary string (e.g. :latest, :stable, a
+	// v-prefixed tag) that need not equal the chart's Chart.yaml version, so
+	// comparing them would report perpetual drift. A tag change is a URL change
+	// and is caught by the url-hash label below.
+	if in.Chart.URL == "" && versionDrifted(in.Chart.Version, ocm.Version) {
```
erhancagirici: Floating/tagless URL refs now have no escape hatch:

  Late-init is disabled in URL mode and the version check is skipped there, so chart.version is fully inert in URL mode. For oci://reg/chart (no tag) or :latest, url-hash always matches and the
  digest label is always "" → permanently up-to-date. Bumping chart.version used to force the re-pull via PullAndLoadChart's case v == "" cache-bypass branch. If intentional, the
  ChartSpec.Version doc comment needs a URL-mode carve-out, it still promises late-initialization unconditionally.

pkg/clients/helm/labels.go
```diff
@@ -0,0 +1,115 @@
...
+// (e.g. a digest with an algorithm prefix longer than "sha256") degrades to
+// "no label" rather than producing a value the API server rejects at deploy.
+func asLabelValue(v string) string {
+	if len(validation.IsValidLabelValue(v)) != 0 {
+		return ""
+	}
+	return v
+}
+
+// EncodeDigestLabel converts a chart digest ("sha256:<64 hex>") into a valid
```
erhancagirici: doc nit:

EncodeDigestLabel / asLabelValue comments now contradict the new semantics
  
Beyond the pre-existing over-long/truncate mismatch, asLabelValue's comment still says it degrades to "no label" — but after dcb7107, "" is a written label meaning "not digest-pinned". An
  unencodable digest embedded in an OCI URL is therefore recorded as unpinned, and the specDigestEnc == "" && in.Chart.Digest != "" guard doesn't cover it (spec digest is empty). Not exploitable
  — the url-hash label catches any such URL change — but reconcile the comments before someone leans on the stale contract.

erhancagirici (commented):
few more nits, and potential follow ups

pkg/controller/cluster/release/release.go
jonasz-lasut: Good catch. This was already on `main` (Connect read the persisted status); fixed here in 1aa4df1 by computing the deploy options before late-init. Covered by `LateInitDoesNotReAdoptOnceOwnershipTaken`.

pkg/controller/cluster/release/release.go
jonasz-lasut: Took your suggestion in 6a95940: the label and the post-deploy status now use `status.ownershipTaken || spec.takeOwnership`. This also stops an upgrade with `takeOwnership` unset from erasing the only record on legacy releases.

pkg/controller/cluster/release/release.go
jonasz-lasut: Confirmed, same on `main`. Tracked in #404.

pkg/controller/cluster/release/observe.go
jonasz-lasut: Intentional, it's what removes the URL-mode late-init loop. A digest pin or an explicit tag is the lever now; documented on `ChartSpec.Version` in f391879, with a floating-tag warning.

pkg/clients/helm/labels.go
jonasz-lasut: Updated in f391879. I also noted at the guard why it can't cover URL-embedded digests: the deploy accepts them, so reporting drift there would upgrade forever.

jonasz-lasut:
Thanks @erhancagirici. Four are fixed here; the legacy digest one predates this PR and is tracked in #404.

erhancagirici (approved):
Thanks a lot @jonasz-lasut for handling this complex feature. LGTM

Issue #202, opened 2023-10-16
Unable to pull Charts from Private Registry


sujith1594:
### What happened?

I'm trying to create a Helm Release using helm-provider, but I'm getting "Certificate Signed by Unknown authority even after adding "insecureSkipTLSVerify: true". Can some one help me how to fix the cert issue or how can I skip the TLS verification?

### How can we reproduce it?

```
apiVersion: helm.crossplane.io/v1beta1
kind: Release
metadata:
  name: helm-testing
spec:
  forProvider:
    insecureSkipTLSVerify: true
    chart:
      pullSecretRef:
        name: helm-secret
        namespace: default
      name: <Chart Name>
      repository: oci://<Private Registry>
      version: 0.1.1
    namespace: default
    values:
      cluster_name: crossplane
  providerConfigRef:
    name: default
```

### What environment did it happen in?
Crossplane version:  1.13.2

Include at least the version or commit of Crossplane you were running. Consider
also including your:

* Cloud provider : AWS
* Kubernetes version (use `kubectl version`): 1.25
* Kubernetes distribution (e.g. Tectonic, GKE, OpenShift) EKS
* OS (e.g. from /etc/os-release) amzn linux

jbw976:
In terms of the `insecureSkipTLSVerify` option, looks like [support for that](https://github.com/helm/helm/commit/08593c8dd6e4b05b1855296f6d7c6247de1d9e6a) was added to [helm v3.12.0](https://github.com/helm/helm/releases/tag/v3.12.0), and we're currently on v3.10.2 in this repo. So perhaps upgrading our dependency would make `insecureSkipTLSVerify` start working 🤔 

Are you interested in testing that out in your environment? 🙏 

There may be further work to make end to end TLS work, but it's hard to tell what may be wrong there without setting up a specific repro environment that we have full control over - otherwise there's too many unknown variables in the environment for which this issue was opened in.

csantanapr:
Thanks for looking into this @jbw976 

So it looks like `insecureSkipTLSVerify` works with http based helm registry, but never worked when using `oci://` registry

@sujith1594 which registry you are using, is it based on an open source and could you provide how is it setup?

sujith1594:
Hi @csantanapr, We are using Redhat Quay which is built on top of Openshift Cluster. Quay is an OCI Registry

cristian98149:
Hi @sujith1594,

which fiels are set in your secret?

[...]

PR #390, opened 2026-08-20
feat: support a custom CA bundle for chart repository/registry TLS

### LINKED ISSUE #391: Support a custom CA bundle for chart repository/registry TLS (not just insecureSkipTLSVerify)

`insecureSkipTLSVerify` is the only option today for a chart repository or OCI registry whose TLS certificate isn't in the base image's system trust store. That's fine when the cert is genuinely untrusted/self-signed and you just want to get unblocked, but it's the wrong tool when the certificate is legitimately signed by a private or internal CA that the operator does trust — you shouldn't have to disable verification entirely just because that CA isn't in the container's default trust store.

This comes up for any private registry (Harbor, Nexus, JFrog, an internal Artifactory, etc.) run behind an org-internal CA rather than a public one.

#202 covers the same underlying pain (`insecureSkipTLSVerify` not working for OCI at the time), and its thread has a workaround comment: mount a CA bundle into the provider pod via `DeploymentRuntimeConfig` + `SSL_CERT_FILE`. That works, but:
- It's provider-pod-wide, not per-`Release` — every `Release` on that provider install is forced to trust the same bundle, even ones pointed at totally unrelated registries.
- It's undocumented outside a GitHub issue comment; nothing in the CRD schema, README, or examples mentions it.
- It only works because of how the underlying HTTP client happens to read that env var — not something the provider actually configures or guarantees.

### What are you expecting to happen?

A field on the `Release` spec that lets you point at a Secret or ConfigMap key holding a PEM-encoded CA bundle, and have the provider trust it for that `Release`'s chart pull — for both classic HTTP(S) chart repositories and OCI registries — without touching `insecureSkipTLSVerify` or any provider-pod-wide config.

### How can we reproduce it?

1. Stand up a chart repository or OCI registry with a certificate signed by a private CA (not in the system trust store).
2. Create a `Release` pointing at it.
3. Observe `x509: certificate signed by unknown authority` on every reconcile, with no way to resolve it short of `insecureSkipTLSVerify: true` (all-or-nothing) or the provider-pod-wide `DeploymentRuntimeConfig` workaround from #202.

barunavo:
Closes #391

## Summary

Adds `spec.forProvider.caBundle` to the `Release` CR (both cluster and namespaced), for trusting a private/internal CA when pulling charts from a chart repository or OCI registry — without disabling TLS verification entirely.

```yaml
spec:
  forProvider:
    caBundle:
      configMapKeyRef:
        name: internal-ca-bundle
        namespace: wordpress   # cluster-scoped Release only
      # key: ca.crt is the default if omitted
```

Reuses the existing `ValueFromSource` type (already used by `valuesFrom`/`patchesFrom`), so it works with either a ConfigMap or a Secret key — no new API type introduced.

## Why

`insecureSkipTLSVerify` (added for #202) skips verification entirely — not appropriate when the registry's cert is legitimately signed by a CA the operator trusts, just not one in the base image's system trust store. The only current workaround is mounting a CA bundle into the provider pod via `DeploymentRuntimeConfig` + `SSL_CERT_FILE` ([documented as a comment on #202](https://github.com/crossplane-contrib/provider-helm/issues/202#issuecomment-2004847589), not in any official doc) — that's provider-pod-wide, not per-`Release`, and only works at all because of how the Go runtime happens to read that env var.

## Implementation note

This needed two separate code paths, not one — traced through `helm.sh/helm/v4`'s source to confirm both:
- **Classic HTTP(S) chart repos**: `action.Pull`/`Install`/`Upgrade` (via embedded `ChartPathOptions`) already have a `CaFile` field that Helm's own getter/repo client reads (`getter.WithTLSClientConfig`). This PR just writes the bundle to a temp file and sets it.
- **OCI registries**: `action.Pull.Run` does **not** use `CaFile` for OCI refs at all — it delegates entirely to `actionConfig.RegistryClient` via `getter.WithRegistryClient`. That registry client is constructed once in `NewClient` via `registry.NewClient()` with no TLS options today. This PR builds a `*http.Client` whose `RootCAs` includes the bundle and passes it via `registry.ClientOptHTTPClient`.

Both are needed for `caBundle` to actually cover both protocols — happy to split into two PRs if that's easier to review, they're independent.

## Test plan
- [x] `go build ./...`, `go vet ./...` — full repo
- [x] `go test ./...` — all pass
- [x] `golangci-lint run ./...` — 0 issues
- [x] `gofmt -l` — clean
- [x] New tests in `pkg/clients/helm/client_ca_bundle_test.go`:
  - `TestHTTPClientTrustingCABundle` — real TLS handshake against a self-signed CA: default client rejects it, the CA-bundle client accepts it, invalid PEM errors.
  - `TestWriteCABundleToTempFile` — content round-trip, uniqueness across calls.
  - `TestNewClient_CABundleWiring` — white-box check that `Args.CABundle` populates `CaFile` on all three action structs correctly, and stays empty (no behavior change) when unset.
  - `TestOCIRegistryPull_WithCABundle` — end-to-end: pushes a real chart to an in-process OCI registry (`google/go-containerregistry`, already a dependency here) served over TLS with a self-signed CA, then pulls it back via Helm's own `registry.Client` — fails without the CA bundle with a genuine x509 error, succeeds with it.
- [x] Updated `examples/{cluster,namespaced}/sample/release-oci.yaml` with a commented usage example
- [x] CRDs regenerated via `go:generate` (controller-gen) — diff isolated to the new field


jonasz-lasut (changes requested):
Hi @barunavo, thank you for the PR - and especially for the implementation note. I checked both of its claims against the `helm.sh/helm/v4 v4.2.3` source and they hold: the classic-repo path reads `ChartPathOptions.CaFile` via [`getter.WithTLSClientConfig` (pull.go#L84)](https://github.com/helm/helm/blob/v4.2.3/pkg/action/pull.go#L84), and OCI pulls delegate entirely to [`p.cfg.RegistryClient` (pull.go#L94-L97)](https://github.com/helm/helm/blob/v4.2.3/pkg/action/pull.go#L94-L97), so the two-path split is correct and both halves are needed. The end-to-end OCI test against a real TLS registry is a nice addition to confirm this behavior, I think we should unify it into E2E suite at some point in time but that's something I'll try to tackle.

There are three changes I'd like to see before merge, plus two smaller notes.

## The temp CA file accumulates unboundedly - one file per reconcile

`writeCABundleToTempFile` runs inside `NewClient` (`pkg/clients/helm/client.go:135`), `NewClient` runs inside `connector.Connect`, and the managed reconciler calls `Connect` on every reconcile. Each reconcile of a `Release` with `caBundle` set therefore creates a new `/tmp/helm-ca-bundle-*.pem` that nothing ever deletes - roughly 144 files/day per `Release` at the default poll interval, for the lifetime of the pod.

The comment's analogy to the chart cache doesn't hold: `/tmp/charts` is a *fixed, keyed* path that gets **reused** across reconciles, while `os.CreateTemp` creates a *new unique* file per call that is never **removed**.

Suggested fix - make the path content-addressed instead of unique:

- Write to `/tmp/ca-bundles/<sha256(content)>.pem`: write to a temp name in the same dir, then `os.Rename` into place (atomic, safe under concurrent reconciles). If the file already exists, just bump its mtime.
- On each `NewClient`, opportunistically delete entries whose mtime is older than an idle TTL (say 24h). No background goroutine needed; the TTL is orders of magnitude above a reconcile duration, and a pull that loses its file in the worst case self-heals on the next reconcile.

The file must live on the real filesystem, because helm reads the path itself via `os.ReadFile` ([tlsutil/tls.go#L73](https://github.com/helm/helm/blob/v4.2.3/internal/tlsutil/tls.go#L73)) - so an in-memory fs can't back the store (an `afero.Fs` seam over the OS fs for testability is fine).

## Classic-repo path drops the system trust store - semantics diverge from OCI

When `CaFile` is set, helm builds the getter's cert pool from that file *alone* ([tlsutil/tls.go#L112-L119](https://github.com/helm/helm/blob/v4.2.3/internal/tlsutil/tls.go#L112-L119)):

```go
if len(to.caPEMBlock) > 0 {
	cp := x509.NewCertPool()
	if !cp.AppendCertsFromPEM(to.caPEMBlock) {
		return nil, errors.New("failed to append certificates from pem block")
	}

	config.RootCAs = cp
}
```

So with `caBundle` set, OCI trusts *system roots + bundle* (your `httpClientTrustingCABundle`), while classic repos trust the *bundle only*. The `Args.CABundle` doc comment ("in addition to the system trust store") is wrong for the classic path. This bites in a real scenario: a mirrored repo whose `index.yaml` sits behind the internal CA but whose `urls:` entries point at a public-CA CDN host.

What does work: concatenate the *system bundle file* with the user bundle into the written file. The provider image is `gcr.io/distroless/static`, which ships the Debian CA bundle at `/etc/ssl/certs/ca-certificates.crt`. Suggested lookup order: `SSL_CERT_FILE` if set (matches Go's own root loading, and keeps the [#202 workaround](https://github.com/crossplane-contrib/provider-helm/issues/202#issuecomment-2004847589) composable), then the `crypto/x509` `root_linux.go` candidates, falling back to user-bundle-only plus a log line if none is readable. Content-address the *concatenated* result so an image update that changes the system bundle rolls the file naturally.

If you'd rather keep this PR small, landing bundle-only for classic repos is acceptable *if* the doc comments on `Args.CABundle` and the CRD field say so explicitly.

## A deleted caBundle source blocks Release deletion

`caBundle` resolution happens in `Connect` (`pkg/controller/cluster/release/release.go:195-201` and the namespaced twin), and the managed reconciler connects before *every* operation - including uninstall. If the referenced ConfigMap/Secret is deleted before the `Release`, `Connect` fails and the `Release` can never be deleted (stuck finalizer), even though the delete path never pulls a chart - `PullAndLoadChart` is only reached from `deploy` (Create/Update).

The repo already has the pattern for exactly this: `Observe` returns early on deletion "to avoid blocking unnecessarily on missing dependencies" (`release.go:259-266`). Suggested: skip the whole `caBundle` branch in `Connect` when `meta.WasDeleted(cr)`.

## Smaller notes

- **Doubled error message:** `NewClient` does `errors.Wrap(err, errFailedToParseCABundle)` around an error that is already `errors.New(errFailedToParseCABundle)`, so a bad bundle surfaces as `failed to parse CA bundle: no PEM certificates found: failed to parse CA bundle: no PEM certificates found`. Drop the wrap or give the two sites distinct messages.
- **`x509.SystemCertPool()` failure is swallowed** (`client.go:244-247`): the fallback to an empty pool silently turns "system roots unavailable" into an inexplicable x509 error on the next OCI pull. A log line at the fallback would make that diagnosable.

Happy to re-review once the leak and the deletion skip are in; the trust-store unification can land here or as an immediate follow-up. Feel free to ping me when you evaluate the review comments.

barunavo:
Thanks for the thorough review, @jonasz-lasut - all three real findings plus both smaller ones are addressed in the latest commit.

1. **Temp file leak** - fixed as suggested: `writeCABundleToFile` now keys the path by the SHA-256 of its own content under `/tmp/ca-bundles/`, so repeated reconciles with the same `CABundle` (the common case) reuse one file instead of creating a new one every time. Each call also opportunistically sweeps entries older than a 24h TTL - no background goroutine.

2. **Trust-store parity** - went with the full fix rather than the narrower doc-only option: `NewClient` now reads a system CA bundle (checking `$SSL_CERT_FILE` first, then the same candidate paths `crypto/x509`'s own `root_linux.go` checks, since there's no public API to get the raw PEM bytes `x509.SystemCertPool()` used) and concatenates it with the supplied `CABundle` before writing `CaFile`. Added an end-to-end test using helm's own `getter.HTTPGetter` against a real TLS server: a bundle unrelated to the server's issuing CA fails alone, succeeds once concatenated with the real one - proving the fix through helm's actual `tlsutil` code path, not just my own assumptions about it.

3. **Stuck deletion** - both `Connect` implementations now skip the `CABundle` branch when `meta.WasDeleted(cr)`, matching `Observe`'s existing pattern. Added a test for each (cluster and namespaced) with a `CABundle` pointing at a Secret that doesn't exist and a `DeletionTimestamp` set - confirmed each fails without the fix by reverting it locally, watching the test fail, then restoring.

4. **Doubled error message** - dropped the redundant wrap at the `NewClient` call site.

5. **Swallowed `SystemCertPool()` failure** - now logs a line before falling back to an empty pool.

Also unified `httpClientTrustingCABundle`'s signature to take a logger (needed for #5), which touched all its existing call sites in tests.

Full suite (`go build`, `go vet`, `go test ./...`, `golangci-lint run`) is clean. Happy to split the trust-store fix into its own commit/PR if that's easier to review separately - let me know.

jonasz-lasut (approved):
Thank you @barunavo LGTM! We really appreciate your contribution

PR #399, opened 2026-09-08
Consume crossplane-runtime #1113

jonasz-lasut:
## What

Bumps `crossplane-runtime` to the latest `main` commit to consume
[crossplane-runtime#1113](https://github.com/crossplane/crossplane-runtime/pull/1113),
which adds a PCU finalizer.

## Why this also touches `pkg/controller/cluster/release/release.go`

crossplane-runtime#1113 turns ProviderConfig deletion protection on by default:
`defaultProviderConfigUsageCleaner` in `pkg/reconciler/managed/reconciler.go`
scans the manager's scheme for the single `ProviderConfigUsage` kind of the
managed resource's scope and, when it finds none or several, falls back to
`NewNopProviderConfigUsageCleaner()` without any log or event.

provider-helm hits that fallback for cluster scope. `apis/cluster/helm.go`
registers both `v1alpha1.ProviderConfigUsage` and `v1beta1.ProviderConfigUsage`,
so with the dependency bump alone the cluster-scoped `Release` reconciler would
never add the usage finalizer and a `ProviderConfig` would stay deletable while
a `Release` is still terminating. The namespaced scheme registers a single usage
kind, so namespaced `Release`s are protected by the bump alone. Running the same
scan the runtime does against `apis/cluster.AddToScheme` and
`apis/namespaced.AddToScheme` at this PR's runtime commit
(`v2.5.0-rc.0.0.20260908074656-9b2fb6b1d1ff`):

```
legacy (cluster-scoped MRs): 2 [apis/cluster/v1alpha1.ProviderConfigUsage apis/cluster/v1beta1.ProviderConfigUsage] -> NO-OP (protection off)
typed  (namespaced MRs):     1 [apis/namespaced/v1beta1.ProviderConfigUsage] -> ACTIVE
```

The second commit passes the connector's `LegacyProviderConfigUsageTracker` to
the reconciler through `managed.WithProviderConfigUsageCleaner`, the override
the runtime exposes for exactly this case. Dropping `v1alpha1` from the scheme
would also resolve it, but that is a deprecation decision for a separate PR.


pkg/controller/cluster/release/release.go
```diff
@@ -88,14 +88,20 @@ const (
...
 	reconcilerOptions := []managed.ReconcilerOption{
 		managed.WithExternalConnector(&connector{
 			client:          mgr.GetClient(),
 			logger:          o.Logger,
-			usage:           resource.NewLegacyProviderConfigUsageTracker(mgr.GetClient(), &helmv1beta1.ProviderConfigUsage{}),
+			usage:           usage,
 			clientBuilder:   kubeclient.NewIdentityAwareBuilder(mgr.GetClient()),
 			newHelmClientFn: helmClient.NewClient,
 		}),
+		managed.WithProviderConfigUsageCleaner(usage),
```
ulucinar, line 104: Making the explicit call is good & preferred. As a follow-up, we will probably remove `ProviderConfig/Usage.v1alpha1` from the manager's scheme.

ulucinar (approved):
Thank you @jonasz-lasut, lgtm.

Issue #213, opened 2024-03-07
Drop v1alpha1 or add conversion webhooks


jcooklin:
### What problem are you facing?

When debugging a composition I was getting the following error.

```cannot compose resources: cannot apply composed resource "spotinstance-controller": failed to create typed patch object (/jc01-fqjml-5m4c8; helm.crossplane.io/v1alpha1, Kind=Release): .spec.forProvider.skipCreateNamespace: field not declared in schema```

This was confusing since the CRD on the cluster clearly had the field.  What I did not immediately recognize was that we were using v1alpha1 in the patch and transform not v1beta1.

@haarchri thanks for your keen 👁️   

### How could Crossplane help solve your problem?

Drop v1alpha1

**Comment by haarchri** [comment]
We faced a problem with the provider-helm in one environment, we using a few old v1alpha1 releases.
In the past we introduced v1beta1 releases and later introduced new fields like `skipCreateNamespace`, which were not implemented into v1alpha1. This has become now an issue issue since v1alpha1 releases are still in etcd. We updated the resources to v1beta1 in their composition, but an automatic storage version migration in etcd did not happen alone ...
As a result, we have now facing an error: `failed to convert new object: .spec.forProvider.skipCreateNamespace: field not declared in schema` Do we have plans to remove v1alpha1 and add a storage version migration like in crossplane ? https://github.com/crossplane/crossplane/pull/4402/ cc: @turkenh

**Comment by haarchri** [comment]
```
kubectl get --raw /apis/helm.crossplane.io/v1alpha1/releases/xyz | jq
{
  "apiVersion": "helm.crossplane.io/v1alpha1",
  "kind": "Release",
  "metadata": {
    "annotations": {
      "crossplane.io/composition-resource-name": "xyz",
      "crossplane.io/external-create-pending": "2024-03-12T22:29:06Z",
      "crossplane.io/external-create-succeeded": "2024-03-12T22:29:08Z",
      "crossplane.io/external-name": "xyz"
    },
    "creationTimestamp": "2024-03-12T22:29:06Z",
    "finalizers": [
      "finalizer.managedresource.crossplane.io"
    ],

```

turkenh:
IIUC, the problem here is different than what https://github.com/crossplane/crossplane/pull/4402 solves. We bumped from v1alpha1 to v1beta1 and marked v1beta1 as the storage version and have both versions in the version list different than what was happening on Crossplane side.

[...]

PR #400, opened 2026-09-10
Apply MR API versioning policy, deprecate `v1alpha1`, add `Release` conversion, lint examples

sergenyalcin:
### Summary

This PR starts applying our API versioning policy to provider-helm's APIs. It deprecates the legacy `v1alpha1` versions of `Release`, `ProviderConfig`, and `ProviderConfigUsage`, adds a real conversion webhook for `Release` (the only one of the three whose `v1beta1` schema actually diverges from `v1alpha1` in ways that need it), and wires up an example-manifest linter to catch schema drift going forward.

#### Deprecation
- `Release`, `ProviderConfig`, and `ProviderConfigUsage` `v1alpha1` now carry `+kubebuilder:deprecatedversion` markers and doc comments pointing at their `v1beta1` replacements.

#### Conversion
- `Release`: `v1beta1` is the `Hub`. Shared fields are copied via an unstructured round-trip rather than a hand-maintained field list, so future additive schema changes don't silently go unhandled. Fields that exist only in `v1beta1` are stashed in a `helm.crossplane.io/release-conversion-data` annotation when converting down to `v1alpha1`, and restored when converting back up — this prevents a write made through the deprecated `v1alpha1` API from silently zeroing out fields it has no way to represent. Verified with a fuzz-based round-trip test (`conversion_test.go`) covering both directions.
- `ProviderConfig`/`ProviderConfigUsage`: deliberately no conversion webhook. Their `v1alpha1`/`v1beta1` schemas are compatible (no renamed/retyped fields) that the CRD's default strategy: None is sufficient; adding a webhook here would be pure overhead.

### Example manifest linting
- New `example-lint` Makefile target (via `kubectl-validate`) validates our example manifests against the CRDs we actually ship, and a matching `example-lint` CI job.

I have:

Tested locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,28 @@ jobs:
       - name: Check Diff
         run: make check-diff
 
+  example-lint:
+    runs-on: ubuntu-latest
+    needs: detect-noop
+    if: needs.detect-noop.outputs.noop != 'true'
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          submodules: true
+
+      - name: Setup Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: go.mod
+
+      - name: Vendor Dependencies
+        run: make vendor vendor.check
+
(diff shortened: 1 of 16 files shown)


jonasz-lasut (approved):
Thank you for this effort @sergenyalcin LGTM!

jonasz-lasut:
I think it also fixes #213 by adding conversion webhooks and long-term by removing v1alpha1 in the next major release

